PR #258, opened 2024-07-17
fix: check gh runner args as nil in case not initialized

### LINKED ISSUE #257: gh runner args not initialized return panic error
We are hitting a panic error when args for gh runner are not initialized

adrianriobo:
Fix #257


pkg/util/ghactions/runner.go
```diff
@@ -12,22 +13,33 @@ type RunnerArgs struct {
...
+		`      ./config.sh --token %s --url %s --name %s --unattended --replace` + "\n" +
+		`      sudo ./svc.sh install` + "\n" +
+		`      chcon system_u:object_r:usr_t:s0 $(pwd)/runsvc.sh` + "\n" +
+		`      sudo ./svc.sh start`
+)
 
-func getRunnerDownloadURLLinux() string {
-	return fmt.Sprintf(runnerBaseURLLinux, runnerVersion, runnerVersion)
-}
+var args *RunnerArgs
```
anjannath, line 42: if we set this to `var args *RunnerArgs = nil` then it will fix the bug we are getting because of nil pointer dereference

pkg/util/ghactions/runner.go
```diff
@@ -42,43 +54,23 @@ func InitGHRunnerArgs(token, name, repoURL string) error {
 }
 
 func GetToken() string {
-	if (args == &RunnerArgs{}) {
-		return ""
-	}
-	return args.Token
+	return util.If(args != nil,
+		args.Token,
+		"")
```
anjannath, line 59: maybe this also needs to be using `if (args != nil)` ?

adrianriobo: yeah will update shortly, thx for catching it

adrianriobo: done

adrianriobo: well..actually created a new ternary op supporting nillable cases

anjannath (approved):

PR #282, opened 2024-08-29
ci: add github actions workflow to test self hosted runner

anjannath:
- [x] need to use a cloud storage for the backend, to be able to remove the instance after the job
- [x] Add `mapt destroy` job to remove provisioned instance

### DIFF
--- a/.github/workflows/provision-hosted-runner.yaml
+++ b/.github/workflows/provision-hosted-runner.yaml
@@ -1,0 +1,62 @@
+name: provision-hosted-runner
+
+on:
+  workflow_call:
+    inputs:
+      operating_system:
+        required: true
+        type: string
+      runner_repo:
+        required: true
+        type: string
+
+jobs:
+  provision_runner:
+    name: provision-runner
+    runs-on: ubuntu-24.04
+    steps:
+      - name: fetch token from API
+        id: fetch_token
+        run: |
+          curl -s -L \
+            -X POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{secrets.GH_PAT_TOKEN}}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/${{github.repository}}/actions/runners/registration-token > token
+          token=$(cat token | jq .token)
+          echo "::add-mask::$token"
+          echo "runner_token=$token" >> "$GITHUB_OUTPUT"
+
+      - name: Download mapt image from artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: mapt
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ github.token }}
+
+      - name: Import mapt image
+        run: |
+          podman load -i mapt.tar
+          podman images
+
+      - name: Run mapt
+        run: |
+          MAPT_IMAGE=$(cat mapt-image)
+          podman run --name mapt-create --rm \
+            -v ${PWD}:/workspace:z \
+            -e ARM_CLIENT_ID=${{secrets.ARM_CLIENT_ID}} \
+            -e ARM_CLIENT_SECRET=${{secrets.ARM_CLIENT_SECRET}} \
+            -e ARM_TENANT_ID=${{secrets.ARM_TENANT_ID}} \
+            -e ARM_SUBSCRIPTION_ID=${{secrets.ARM_SUBSCRIPTION_ID}} \
+            -e AZURE_STORAGE_ACCOUNT=${{secrets.AZURE_STORAGE_ACCOUNT}} \
+            -e AZURE_STORAGE_KEY=${{secrets.AZURE_STORAGE_KEY}} \
+            ${MAPT_IMAGE} azure ${{inputs.operating_system}} create \
+              --spot --project-name "az-ghrunner" --conn-details-output /workspace \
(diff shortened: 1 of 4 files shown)


.github/workflows/hosted-runner.yaml
```diff
@@ -0,0 +1,93 @@
+name: Test Self hosted runner addition
+
+on:
+  workflow_call:
```
adrianriobo, line 4: We will need a strategy based on workflow_run to get access to secrets and test the binary build from the forked repo, with the current build you have here it will build it form the head of the main instead of from the PR

adrianriobo: Similar to https://github.com/crc-org/ci-definitions/blob/70d016f5fd15e7cbdf8e3056c500654b0196a049/.github/workflows/crc-builder-pusher.yml#L18 stragtey for build / push

anjannath: we also have to store the backend in a cloud storage to be able to destroy the created resources later

adrianriobo: @anjannath  ^^ Just added all required secrets for this:

- ARM_TENANT_ID
- ARM_SUBSCRIPTION_ID
- ARM_CLIENT_ID
- ARM_CLIENT_SECRET
- AZURE_STORAGE_ACCOUNT
- AZURE_STORAGE_KEY

Also to use the remote storage for backed-url the url would be something like (notice base name should exactly match `mapt-gh-runner-mapt-state`):

```
--backed-url azblob://mapt-gh-runner-mapt-state/${{ github.repository }}-${{ github.run_id }} 
```

anjannath: It seems `workflow_run` only works when the workflow is already in the `main` branch, so need to test this first on my fork and then update the PR, we'll have to initially merge something and then create follow up PR to fix if any issues found

adrianriobo:
Hey can we test  selfhosted runner running crc on it. I am curious on the user running the agent. 

I mean when I set up ssh I had to trick the user starting the service, as with default setup it is a system user (privileged) as so crc is not working. How is working with the gh agent?

anjannath:
> Hey can we test selfhosted runner running crc on it. I am curious on the user running the agent.
> 
> I mean when I set up ssh I had to trick the user starting the service, as with default setup it is a system user (privileged) as so crc is not working. How is working with the gh agent?

the runner agent running as service using the `nt authority\network service` account, i did a test run here: https://github.com/anjannath/mapt/actions/runs/11459558776/job/31884314010

will have to try running `crc start` to see it it works with this user, the `NetworkService` is also a privileged user account i think

anjannath:
@adrianriobo the `workflow_call` strategy will pick the commit from the PR if we add a step for `checkout code` in the workflow job

so currently this PR adds to re-usable workflows, `provision_hosted_runner` and `destroy_hosted_runner` and a `build_on_hosted_runner` workflow that makes use of the reusable workflows, but currently it seems the secrets are not accessible to the workflows as they are not part of the repo main branch yet

the build-on-hosted-runner workflow additionally has to fetch a runner registration token which is needed to add a self hosted runner and is an input for the provision_hosted_runner workflow

anjannath:
@adrianriobo this is finally ready for review again, got some successful runs on my fork, https://github.com/anjannath/mapt/actions/runs/11698979004

adrianriobo (approved):
^^ @anjannath  About your comment you can add a filter on PRs were any file of the util/ghactions is modified i.e. https://github.com/crc-org/ci-definitions/blob/main/.github/workflows/crc-builder-builder.yaml#L6C3-L8C79 

Beyond that LGTM

.github/workflows/provision-hosted-runner.yaml
```diff
@@ -0,0 +1,62 @@
...
+            -e ARM_SUBSCRIPTION_ID=${{secrets.ARM_SUBSCRIPTION_ID}} \
+            -e AZURE_STORAGE_ACCOUNT=${{secrets.AZURE_STORAGE_ACCOUNT}} \
+            -e AZURE_STORAGE_KEY=${{secrets.AZURE_STORAGE_KEY}} \
+            ${MAPT_IMAGE} azure ${{inputs.operating_system}} create \
+              --spot --project-name "az-ghrunner" --conn-details-output /workspace \
+              --backed-url "azblob://mapt-gh-runner-mapt-state/${{ github.repository }}-${{ github.run_id }}" \
+              --install-ghactions-runner --ghactions-runner-name "az-runner-${{inputs.operating_system}}-${{github.event.workflow_run.id}}" \
+              --ghactions-runner-repo "${{inputs.runner_repo}}" --ghactions-runner-token ${{steps.fetch_token.outputs.runner_token}}
+
+      - name: wait for runner to start accepting jobs
```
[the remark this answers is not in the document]
anjannath: during testing i found that when the job executed immediately the runner would never pick it up, couldn't figure out why this was happening

i just tried opening a second PR just to test by running a different workflow targeting that runner then it'd work and it'd also pickup the pending `test_run_selfhosted_runner` job, so figured that if there's a bit of delay between the runner getting added and the next job getting triggered it all works

anjannath:
> ^^ @anjannath About your comment you can add a filter on PRs were any file of the util/ghactions is modified i.e. https://github.com/crc-org/ci-definitions/blob/main/.github/workflows/crc-builder-builder.yaml#L6C3-L8C79
> 
> Beyond that LGTM

actually this cannot be used, since we trigger the `build-on-hosted-runner` workflow based on if `oci-build` workflow is finished, and then on the job check if the event was a pull_request or not

PR #283, opened 2024-09-02
change tkn image set in makefile

lilyLuLiu:


anjannath:
LGTM

so when we want to now push a released 8.0.0 image we can use, e.g `make VERSION=8.0.0 tkn-push`

tkn/infra-aws-mac.yaml
```diff
@@ -99,7 +99,7 @@ spec:
 
   steps:
   - name: provisioner
-    image: quay.io/redhat-developer/mapt:v0.7.0-dev
+    image: <IMAGE>
```
adrianriobo, line 102: We need to take care of  

`app.kubernetes.io/version: "0.7.0-dev"` 

on the annotations part of the yaml.

This applies to all other yaml files

adrianriobo:
Also I notice the `tkn-bundle/tkn-check` passed but that should not be the case, as the spec do not contain a valid image on it, it just contains the `<IMAGE>` field as the tkn-template is not being used there.

We have to issues there:

- Ensure the template function is used somehow before they are deployed to the cluster on the check
- Look for a way to dry/run or some type of test which may does not require credentials / or something we can pass empty  to test the task will be able to run.

tkn/infra-aws-mac.yaml
lilyLuLiu: sed command also covered the label now.

adrianriobo: :+1:

lilyLuLiu:
@adrianriobo , I keep the orignal tkn yaml files and use another folder to store the templates. So `tkn-bundle/tkn-check` can check the validation of the tkn file. Also add `make tkn-update` in makefile for updating the version for tkn.

adrianriobo (approved):

PR #277, opened 2024-08-12
convert vm specs to provider specific instance type name

anjannath:
fixes #281

PR #284, opened 2024-09-02
Spot module as a separate package

anjannath:
depends on #277


pkg/provider/util/instancetypes/instancetypes.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func validate(cpus, memory int32, arch arch) error {
+	if cpus > 0 && memory > 0 && arch.String() != "" {
+		return nil
+	}
+	return fmt.Errorf("Invalid values for CPUs: %d, Memory: %d and Arch: %s", cpus, memory, arch)
+}
+
+func (r *AwsInstanceRequest) GetMachineTypes() ([]string, error) {
```
adrianriobo, line 53: Do you need to rebase and adapt for this?

adrianriobo:
Also same comment as per type of machine applies here, if we rely on this module; is it only SDKs dependent?

pkg/provider/util/instancetypes/instancetypes.go
anjannath: it is rebased and ready for review now :)

anjannath:
> Also same comment as per type of machine applies here, if we rely on this module; is it only SDKs dependent?

yes, this also only depends on the sdks, i've added example on how to use it in: b553996b7071adfcc624d627320557eaa49e97ea (not sure we should merge this or not, i am okay to drop it from the PR if needed)

examples/spotprice/main.go
```diff
@@ -0,0 +1,55 @@
...
+		MemoryGib:  8,
+		Arch:       "amd64",
+		NestedVirt: false,
+		Os:         "fedora",
+		OSVersion:  "39",
+	}
+
+	// Get the lowest price for only aws
+	//spi, err := spotReq.GetAwsLowestPrice()
+
```
adrianriobo, line 34: I guess this is expecting to have the specific Envs with credentials ...can we add that here? I mean os.SetEnv(AWS..., SampleSecretKey)....

To make more clear how we are expecting the credentials to be set?

anjannath: updated with the required env variables that needs to be set for both aws and azure

anjannath:
> Also I just merged the defaults so probably this will need rebase again ?

thanks, rebased on main

PR #298, opened 2024-09-27
add azure fedora support

anjannath:


pkg/provider/azure/module/virtual-machine/virtual-machine.go
```diff
@@ -38,6 +40,18 @@ type VirtualMachineRequest struct {
 // Create virtual machine based on request + export to context
 // adminusername and adminuserpassword
 func (r *VirtualMachineRequest) Create(ctx *pulumi.Context) (*compute.VirtualMachine, error) {
+	var imageReferenceArgs compute.ImageReferenceArgs
+	if r.ImageID != "" {
```
adrianriobo, line 44: len(r.ImageID) > 0 ?

pkg/provider/azure/action/linux/imageref.go
```diff
@@ -60,7 +62,24 @@ func getImageRef(osTarget OSType, arch string, version string) (*imageReference,
 			offer:     ir.offer,
 			sku:       fmt.Sprintf(ir.sku, versions[0], versions[1]),
 		}, nil
+	case Fedora:
+		return &imageReference{}, nil
 	}
 	return nil, fmt.Errorf("os type not supported")
 
 }
+
```
adrianriobo, line 71: Is it possible to unify the struct I was using and this other way the ImageID name formatted?

anjannath: we can add another field to the `imageReference` struct:
```go
type imageReference struct {
	publisher string
	offer     string
	sku       string
        // community gallery image ID
        id string
}

```

which would be empty for the existing images, but will have a value for `Fedora` and then while using it in `virtual-machine.go` we could check for `len(imageReferece.id) > 0` and use only the `id` if it exists, otherwise the existing `publisher`, `offer` and `sku` is used for selecting the image

adrianriobo (commented):
Need the tekton task and the link to this info on the main README.md table

pkg/provider/azure/action/linux/imageref.go
adrianriobo: Great!!

anjannath:
> Need the tekton task and the link to this info on the main README.md table

Added the tekton task, but its not tested, i copied the `tkn/infra-azure-rhel.yaml` and basically changed rhel to fedora :) but also introduced the `cpus` `memory` params instead of `vm-size`

Issue #311, opened 2024-10-15
[Enhancement] Enable multi-arch build for mapt


adrianriobo:
Currently mapt and oci container is built only for x86 it should be built for multi-arch

adrianriobo:
Here there is a sample on how we manage this on other project:

Makefile for build and aggreagate to manifest
https://github.com/crc-org/ci-definitions/blob/71d9c0cc3371aec456fbb5c048cb8c6cd52c8e0a/Makefile#L66

Containerfile with target arch arg to manage the internal build
https://github.com/crc-org/ci-definitions/blob/main/crc-builder/oci/Containerfile.linux#L4

adrianriobo:
Hey @lilyLuLiu I actually open an issue https://github.com/crc-org/ci-definitions/issues/32 as it seems this is not working as expected...mainly due to the strategy required for forked repos.

We basically create those images on the forked repo (as long as they are not tagged versions) then save as tar upload as artifacts and from the upstream context load and push to internal ghcr. 

But it seems that save operation does not allow us to do the multiarch 

https://github.com/containers/podman/issues/22882

adrianriobo:
To make the builded pusher strategy works there is a tricky thing needed to make in order to save images you can see https://github.com/crc-org/ci-definitions/pull/34/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R72

adrianriobo:
Closed by #318

PR #318, opened 2024-10-21
Enable multi-arch build 

lilyLuLiu:
https://github.com/redhat-developer/mapt/issues/311


lilyLuLiu:
@adrianriobo There is no arm64 machine in github action runner. So we can only use action for amd64 mapt image build and upload.

adrianriobo (approved):
LGTM

PR #323, opened 2024-10-30
bump tkn to mapt:v0.8.0-dev

lilyLuLiu:


adrianriobo:
Hey @lilyLuLiu just saw this makes sense to use the latest so we do not need to create a new version for mapt immediately. In this PR you just changed the task for fedora..can you do it for all tasks?  Probably best way is to run `make tkn_update` so all versions will be changed at once + version annotation which is actually missing here

adrianriobo (commented):
Hey @lilyLuLiu just saw this makes sense to use the latest so we do not need to create a new version for mapt immediately. In this PR you just changed the task for fedora..can you do it for all tasks? Probably best way is to run make tkn_update so all versions will be changed at once + version annotation which is actually missing here

adrianriobo (approved):

PR #340, opened 2024-11-28
ci: build image for self hosted gh runner test

anjannath:
this workflow uses path filters and runs only when files related to self hosted runner code changes

it then triggers the workflow for the  self hosted runner test, making the workflow run only for file changes in the self hosted runner modules

### DIFF
--- a/.github/workflows/build-img-ghrunner-test.yaml
+++ b/.github/workflows/build-img-ghrunner-test.yaml
@@ -1,0 +1,43 @@
+name: build-img-ghrunner-test
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths:
+      - 'pkg/util/ghactions/*.go'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-img-ghrunner-test.yaml
```diff
@@ -0,0 +1,50 @@
+name: build-img-ghrunner-test
+
+on:
+  push:
+    branches: [ main ]
```
adrianriobo, line 5: I think we can remove push, as I set up rule for require rebase on PRs...meaning just before merge we already tested with latest changes in code.  WDYT?

anjannath: yeah we can remove it in that case

adrianriobo (approved):

PR #349, opened 2024-12-12
feat: refactor mac request / release flow

### LINKED ISSUE #347: Move replace root task to release operation

adrianriobo:
This commit move the time consuming effort for having fresh machines on the mac pool from the request action to the release. Now release will manage the root volume replace task and request will create new user credentials and set machine as locked to avoid new workloads to make use of it

Fix #347


cmd/mapt/cmd/aws/hosts/mac.go
```diff
@@ -26,7 +26,7 @@ const (
 	archDefault       string = "m2"
 	osVersion         string = "version"
 	osVersionDesc     string = "macos operating system vestion 11, 12 on x86 and m1/m2; 13, 14 on all archs"
-	osDefault         string = "14"
+	osDefault         string = "15"
```
anjannath, line 29: seem we are duplicating the constants for default values of `arch` and `os` in `pkg/provider/aws/action/mac/constants.go`

adrianriobo: Ahh yeah..may we need to use those from action on the cmd ...as on actions we need them for sure

anjannath (approved):

PR #351, opened 2024-12-13
fix: default workdir with rights for any user

### LINKED ISSUE #350: [bug] Error creating Pulumi.yaml while running mapt on openshift-ci
We are seeing an error while trying to create Pulumi.yaml

````
11 skipped lines...
INFO[2024-12-11T13:07:04Z] DEBU running 'mapt azure aks create'              
DEBU context initialized for mapt245d91ec         
DEBU Creating AKS                                 
DEBU managing stack stackCreateAKS-aks            
push_pin
ERRO Failed to create or select stack: failed to create stack: failed to create workspace, unable to save project settings: open Pulumi.yaml: permission denied 
```

adrianriobo:
Previously workdir was / and only root user was able to write there, when running mapt on an ocp cluster with security context enable and non-root user mapt fail as it requires to create some ephemeral files on the current workdir. This commit will create a path on the image with open rights and set it as workdir to allow create the required files.

Fix #350

### DIFF
--- a/oci/Containerfile
+++ b/oci/Containerfile
@@ -24,17 +24,15 @@ LABEL org.opencontainers.image.authors="Redhat Developer"
 
 COPY --from=builder /workspace/out/mapt /workspace/pulumi/pulumi /usr/local/bin/
 
-ENV PULUMI_CONFIG_PASSPHRASE="passphrase" \
-    AWS_SDK_LOAD_CONFIG=1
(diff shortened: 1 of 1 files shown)


adrianriobo:
@zdrapela can you test this image on openshfit-ci? 

`ghcr.io/redhat-developer/mapt:pr-351` and comment here if it works now

anjannath (approved):

Issue #313, opened 2024-10-15
[BUG] Spot instance type recommendation not available on recommended region


adrianriobo:
There is some checks missing when looking for best spot price / type machine:

We can see: 

```bash
[37mDEBU[0m Based on avg prices for instance types [m4.large m5.large m5a.large m5ad.large m5d.large m5dn.large m5n.large m5zn.large m6a.large m6i.large m6id.large m6idn.large m6in.large m7a.large m7i-flex.large] is az eu-west-2b, current avg price is 0.05 and max price is 0.05 with a score of 9 
[36mINFO[0m @ updating.............                      
[36mINFO[0m  +  rh:qe:aws:bso main-bso-bso creating (0s)  
[36mINFO[0m  +  rh:qe:aws:bso main-bso-bso created       
[36mINFO[0m  +  pulumi:pulumi:Stack debug-fedora-spotOption-debug-fedora created (10s)  
[36mINFO[0m Outputs:                                     
[36mINFO[0m     avg   : 0.0452                           
[36mINFO[0m     az    : "eu-west-2b"                     
[36mINFO[0m     max   : 0.0452                           
[36mINFO[0m     region: "eu-west-2"                      
[36mINFO[0m     score : 9                                
```

But when we try to use the recommended machine type we got:

```bash
Diagnostics:
  aws:ec2:Eip (eip-publicmain-afd-net):
    warning: urn:pulumi:stackFedoraBaremetal-debug-fedora::debug-fedora::aws:ec2/eip:Eip::eip-publicmain-afd-net verification warning: use domain attribute instead

  pulumi:pulumi:Stack (debug-fedora-stackFedoraBaremetal-debug-fedora):
    error: update failed

  aws:autoscaling:Group (main-afd-asg):
    error: 1 error occurred:
    	* creating Auto Scaling Group (main-afd-asg-89ce230): operation error Auto Scaling: CreateAutoScalingGroup, https response error StatusCode: 400, RequestID: 0175d099-0523-431f-9dee-a39975af885d, api error ValidationError: The specified instance type m5zn.large is not valid

Resources:
    + 20 created

Duration: 2m45s
```

adrianriobo:
This is partially fixed with https://github.com/redhat-developer/mapt/pull/314/commits/77281e5c39060077c64f53ca8bf3914bccf09c6b but still there are inconsistencies for all the types of machines and regions (i.e. windows on aws does not using it at all). 

[...]

PR #353, opened 2024-12-16
check if best spot price location supports resource group

anjannath:
it is found that the location suggested by the spot calculation sometimes do not support the creation of Resource Groups  which is a requirement for `mapt` in those cases we use a pre-defined location for creating the Resource group

related issue: #313

### DIFF
--- a/pkg/provider/azure/azure.go
+++ b/pkg/provider/azure/azure.go
@@ -1,6 +1,8 @@
 package azure
 
 import (
+	"slices"
+
 	"github.com/redhat-developer/mapt/pkg/manager"
 	"github.com/redhat-developer/mapt/pkg/manager/credentials"
 )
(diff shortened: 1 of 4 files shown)


pkg/provider/azure/azure.go
```diff
@@ -11,7 +13,55 @@ func GetClouProviderCredentials(fixedCredentials map[string]string) credentials.
 		FixedCredentials:  fixedCredentials}
 }
 
-var DefaultCredentials = GetClouProviderCredentials(nil)
+var (
+	DefaultCredentials               = GetClouProviderCredentials(nil)
+	locationsSupportingResourceGroup = []string{
```
adrianriobo, line 18: DId you check if is there any api call to know region supporting RG?

adrianriobo: Otherwise LGTM

anjannath: i did try to find, but couldn't find any API for getting the list of all locations where resource groups are supported, so went with putting a list of good locations in the code

we also need such an API to handle the missing images issue

adrianriobo: through graph?

adrianriobo (approved):
LGTM

PR #355, opened 2024-12-17
Added debug flag to increase verbosity

### LINKED ISSUE #344: [BUG] `azure aks create --help` suggests using s3 bucket
When running `mapt azure aks create --help`, it suggests using s3 bucket for  `--backed-url`, but it should suggest using `azblob` instead.

Full output:
```
mapt azure aks create --help
Usage:
  mapt azure aks create [flags]

Flags:
      --conn-details-output string       path to export host connection information (host, username and privateKey)
      --enable-app-routing               enable application routing add-on with NGINX
  -h, --help                             help for create
      --location string                  location for created resources in case spot flag (if available) is not passed (default "West US")
      --only-system-pool                 if we do not need bunch of resources we can run only the systempool. More info https://learn.microsoft.com/es-es/azure/aks/use-system-pools?tabs=azure-cli#system-and-user-node-pools
      --spot                             if spot is set the spot prices across all regions will be cheked and machine will be started on best spot option (price / eviction)
      --spot-eviction-tolerance string   if spot is enable we can define the minimum tolerance level of eviction. Allowed value are: lowest, low, medium, high or highest (default "lowest")
      --tags stringToString              tags to add on each resource (--tags name1=value1,name2=value2) (default [])
      --version string                   AKS K8s cluster version (default "1.30")
      --vmsize string                    VMSize to be used on the user pool. Typically this is used to provision spot node pools (default "Standard_D8as_v5")

Global Flags:
      --backed-url string     backed for stack state. Can be a local path with format file:///path/subpath or s3 s3://existing-bucket
      --project-name string   project name to identify the instance of the stack
```

### LINKED ISSUE #354: [enhancement] Added debug flag to enable increased verbosity for pulumi
Currently there is no way to increase verbosity for pulumi in case we need to troubleshoot an issue

adrianriobo:
Fix #354 
Fix #344


anjannath:
LGMT, it'd be good to also have ability to set the log level , currently at the `logLevel = 9` there's a lot of logs

anjannath (approved):

adrianriobo:
Yeah probably make sense I went fully log without checking i.e. level 3 and debug traces...I will extend the PR to add a param for debug-verbosity

adrianriobo:
Added a new DebugLevel flag as per suggestion

anjannath (approved):

PR #362, opened 2024-12-18
configure ntp servers for fedora on aws using cloud-init

### LINKED ISSUE #325: [Bug] Fedora arm64 instance on AWS is booting with ntp out of sync
After boot a Fedora arm64 on AWS trying to install some packages we end up failing with ssl certificate issues due to ntp on the server out of sync.

Check https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configure-ec2-ntp.html

anjannath:
fixes #325


adrianriobo (approved):
LGTM

PR #377, opened 2025-01-13
Mac-pool service

adrianriobo:
Fix #376 

Depends on #352

PR #382, opened 2025-01-23
Add Cirrus support

### LINKED ISSUE #348: [enhancement] Mapt as CIRRUS persistent worker
This feature will enable the usage of provisioned instances as persistent workers for Cirrus CI/CD system

### LINKED ISSUE #376: Refactor mac current actions as modules
In order to implement pool operations we will reuse the functionality offered as actions, due to technical design we would need to spent time on refactor the code as modules functionality for better re usability

### LINKED ISSUE #392: [Bug] Wrong region when destroying serverless
```bash
stderr: error: read ".pulumi/meta.yaml": blob (key ".pulumi/meta.yaml") (code=Unknown): BucketRegionError: incorrect region, the bucket is not in 'eu-west-3' region at endpoint '', bucket is in 'us-east-1' region
```

Main problem is backed url is tied to a specific region, and somehow that information is not being passed when destroying from serverless on a different region. Actually the execution includes an env `AWS_REGION` with current region which is causing the error.

Hint use https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#Client.HeadBucket to get the region for the backedURL bucket and use it when setting credentials on AWS on destroy operation

adrianriobo:
Fix #376 
Fix #348 
Fix #392 

Depends on #377


pkg/integrations/cirrus/snippet-darwin.sh
```diff
@@ -0,0 +1,47 @@
...
+    <key>StandardErrorPath</key>
+    <string>/tmp/com.cirrus.client.err.log</string>
+</dict>
+</plist>
+EOF
+
+sudo mv com.cirrus.client.plist /Library/LaunchDaemons/
+sudo chown root:wheel /Library/LaunchDaemons/com.cirrus.client.plist
+sudo chmod 644 /Library/LaunchDaemons/com.cirrus.client.plist
+sudo launchctl load /Library/LaunchDaemons/com.cirrus.client.plist
```
anjannath, line 47: this is going to run the agent as a Privileged user no? for running as normal user we need to put this in `~/Library/LaunchAgents`

anjannath (approved):

Issue #364, opened 2024-12-18
[enhacement] Speed up gh oci-build action (Move to new gh linux arm64 runners)


adrianriobo:
Currently we build multi-arch images on an x86 ubuntu runner, so the emulation of arm64 to build the arch is taking a lot of time. We can customize a bit the action to speed up things:

* Review Containerfile to ensure the strategy for partial images (images built which can be cached)
* Add cache for images path on the runners
* Split the build in parallel: one runner x86 one runner (emulated) arm64

adrianriobo:
We can make use of new added linux-arm64 runners to avoid emulate arm64 builds which are really time consuming

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

rohanKanojia:
I want to work on this issue. Could you please assign it to me?

adrianriobo:
done!

rohanKanojia:
> emulation of arm64 to build the arch is taking a lot of time 

Is this the step that's consuming a lot of time while building container image? Or do we want to improve the image creation time by building it parallel?

https://github.com/redhat-developer/mapt/blob/35507db350208819daeed4703afa2816c2da8745/oci/Containerfile#L52

I was thinking of splitting the Containerfile into three files:
```
oci
├── Containerfile.amd64
├── Containerfile.arm64
└── Containerfile.base
```
We can move the common parts shared by both platforms in another base container image and then upload it as GitHub artifact. 

`oci-builds` github action would be split into three GitHub actions:
- `oci-build-base` would build common container image
- `oci-build-arm64` would build for arm platform (depend on base workflow)
- `oci-build-amd64` would build for x86_64 (depend on base workflow)


Do you think this approach is okay? Or shall we keep a single `Containerfile` as it's generic enough to handle both platforms?

adrianriobo:
Ahh I think you should keep the Containerfile as it is...what you should focus is on split the Makefile targets and the gh flows to be executed in parallel....

rohanKanojia:
@adrianriobo : Shall we close this issue ?

adrianriobo:
yeah

PR #413, opened 2025-03-06
ci : Move arm64 platform image build to a separate github workflow

rohanKanojia:
# Description

~⚠️ This is still work in progress. I haven't verified that behavior is still same as single target yet.~

Related to #364

Split oci-build Makefile target into different targets for handling amd64 and arm64 platforms separately.

Use a separate github action workflow to invoke separated Makefile targets to create container images

### DIFF
--- a/.github/workflows/build-oci.yaml
+++ b/.github/workflows/build-oci.yaml
@@ -1,3 +1,19 @@
+#
+# Copyright (C) 2025 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
 name: oci-builds
 
 on:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-oci.yaml
```diff
@@ -9,56 +25,70 @@ on:
...
-          name: mapt
-          path: mapt*
+          name: mapt-${{ env.ARCH_TYPE }}
+          path: |
+            mapt-image
+            mapt-${{ env.ARCH_TYPE }}*
 
   push:
     name: push
-    if: github.event_name == 'push'
```
adrianriobo, line 48: I think the `if: github.event_name == 'push' ` in the push action is needed otherwise it will try to push on PR and it fail due to credentials missing...actually this is the current failure you can see on the PR

.github/workflows/push-oci-pr.yml
```diff
@@ -19,18 +35,24 @@ jobs:
...
+        uses: actions/download-artifact@cc203385981b70ca67e1cc392babf9cc229d5806 # v4.1.9
         with:
-          name: mapt
+          name: mapt-amd64
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ github.token }}
+      - name: Download mapt assets
+        uses: actions/download-artifact@cc203385981b70ca67e1cc392babf9cc229d5806 # v4.1.9
+        with:
+          name: mapt-arm64
```
adrianriobo, line 46: I think you can use `pattern` instead so you can download both on the same action https://github.com/actions/download-artifact?tab=readme-ov-file#usage

adrianriobo (approved):
LGTM

rohanKanojia:
Please wait I'm seeing a failure on one of the runs on my fork https://github.com/rohankanojia-forks/mapt/actions/runs/13722854013 . Let me double check it then you can merge.

.github/workflows/build-oci.yaml
```diff
@@ -9,56 +25,67 @@ on:
...
+    needs: build-and-upload
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - name: Download mapt oci flatten images
-        uses: actions/download-artifact@v4
+      - name: Download mapt amd64 oci flatten images
```
adrianriobo, line 82: typo: name contains the arch here

.github/workflows/build-oci.yaml
```diff
@@ -9,56 +25,67 @@ on:
...
-          make oci-save
+          make oci-build-${{ env.ARCH_TYPE }}
+          make oci-save-${{ env.ARCH_TYPE }}
 
       - name: Upload mapt artifacts for PR
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
         with:
-          name: mapt
-          path: mapt*
```
adrianriobo, line 44: I think you can keep this no?

rohanKanojia: As far as I remember I was getting an error in github action that it was trying to upload two artifacts with same name `mapt` . Now we're uploading two different artifacts for different architectures.

adrianriobo: sorry I meant for the `path` name should be as you said 

```
name: mapt-${env.ARCH}
path: mapt*
```

rohanKanojia: Oh, I see. You're right. Let me update it as per your suggestion.

adrianriobo:
Now it seems good no??? really great moving from 40 mins to build to 10 mins :rocket: 

feel free to `rebase and merge` at your discretion I keep the LGTM

rohanKanojia:
okay, on my fork it's passing now https://github.com/rohankanojia-forks/mapt/actions/runs/13723908545/job/38386161379 . 

I think it's okay to merge now. If some issue arises, I'll create a follow up PR to fix it.

PR #422, opened 2025-03-12
ci : Add a check in push workflow to only execute when an artifact is present

### LINKED ISSUE #418: [bug] gh action push not working as expected on PRs
As now the trigger flow contains 2 jobs...it is triggering the push pr 3 times we need a sync mechanism for this....i.e. on the build save for both arm64 and amd64 artifacts should be done on one job...and we should filter for it on the worflow_run

rohanKanojia:
## Description

Fix #418 

Due to split of build amd and arm workflows, push workflow_run is getting triggered multiple times.

+ Add another job that would combine both  image artifacts into one artifact. 
+ In push_pr workflow, add a condition to only execute workflow when both `mapt-amd64.tar` and `mapt-arm64.tar` artifacts are found.

### DIFF
--- a/.github/workflows/build-oci.yaml
+++ b/.github/workflows/build-oci.yaml
@@ -54,20 +54,35 @@ jobs:
         run: |
           make oci-build-${{ env.ARCH_TYPE }}
           make oci-save-${{ env.ARCH_TYPE }}
-          echo ${IMG} > mapt-image
-
       - name: Build and Push image for Release
         if: github.event_name == 'push'
         run: |
           make oci-build-${{ env.ARCH_TYPE }}
           make oci-save-${{ env.ARCH_TYPE }}
-
       - name: Upload mapt artifacts for PR
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: mapt-${{ env.ARCH_TYPE }}
           path: mapt*
-
+  combine-artifacts:
+    name: combine-artifacts
+    needs: build-and-upload
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Download mapt oci flatten images
+        uses: actions/download-artifact@cc203385981b70ca67e1cc392babf9cc229d5806 # v4.1.9
+        with:
+          pattern: mapt-*
+      - name: copy both artifacts into single directory
+        env:
+          IMG: ghcr.io/redhat-developer/mapt:pr-${{ github.event.number }}
+        run: echo ${IMG} > mapt-image
+      - name: Upload combined mapt artifacts for PR
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
+        with:
+          name: mapt-arm64-and-amd64
+          path: mapt-*
   push:
     name: push
     if: github.event_name == 'push'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-oci.yaml
```diff
@@ -67,7 +66,28 @@ jobs:
...
+      - name: copy both artifacts into single directory
+        run: |
+          mkdir mapt-arm64-and-amd64
+          mv mapt-amd64/* mapt-arm64-and-amd64/
+          mv mapt-arm64/* mapt-arm64-and-amd64/
+      - name: Upload combined mapt artifacts for PR
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
+        with:
+          name: mapt-arm64-and-amd64
+          path: mapt-arm64-and-amd64/
```
adrianriobo, line 90: you can use pattern instead of create a folder and added them there...then even the download part would be easier as they would be uncompressed on the current path there and you will have both tars... 

actually you are doing `mapt-amd64/*` b/c inside there is tar and image but image has the same value on both....

so probably you can grab the image value within this job....and remove it from builders....

then you would have something like:

mapt-amd64.tar
mapt-arm64.tar
mapt-image

and you can upload with patter: `mapt-*`

.github/workflows/push-oci-pr.yml
```diff
@@ -25,40 +25,49 @@ on:
...
     if: |
       github.event.workflow_run.conclusion == 'success' &&
       github.event.workflow_run.event == 'pull_request'
     runs-on: ubuntu-24.04
-    permissions:
-      contents: read
-      packages: write
     steps:
-      - name: Download mapt assets
+      - name: Try to Download Artifact
```
adrianriobo, line 34: :) Why try?....just out of curiosity why change the assertion.... (download) for a supposition (try to downlowad) ?

adrianriobo: Ahh b/c there is no way to get the job name to filter it?

.github/workflows/build-oci.yaml
```diff
@@ -67,7 +66,28 @@ jobs:
         with:
           name: mapt-${{ env.ARCH_TYPE }}
           path: mapt*
+  combine-artifacts:
+    name: combine-artifacts
```
adrianriobo, line 70: May you can add a condition to combine only on pull_request....as for push you directly push to quay

adrianriobo:
Hey can you take care of this https://github.com/redhat-developer/mapt/actions/runs/13893232540/job/38868595571?pr=410 in this one...I think this is also a regression with the previous change on the gh actions / makefile ..or feel free to add as another following PR

rohanKanojia:
@adrianriobo : Apologies for the delay, let me try to address these comments.

adrianriobo:
no worries, 

just go over it when you have time, no rush

.github/workflows/build-oci.yaml
```diff
@@ -54,20 +54,37 @@ jobs:
...
           name: mapt-${{ env.ARCH_TYPE }}
           path: mapt*
-
+  combine-artifacts:
+    name: combine-artifacts
+    needs: build-and-upload
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout code
```
adrianriobo, line 73: Why do you need checkout here?

adrianriobo (approved):
LGTM

PR #450, opened 2025-03-24
ci : fix build-img-ghrunner-test pipeline as per image arch split

rohanKanojia:
## Description
Related to https://github.com/redhat-developer/mapt/pull/422#issuecomment-2728497254

Fix failure in `build-img-ghrunner-test` github action due to oci build target split

### DIFF
--- a/.github/workflows/build-img-ghrunner-test.yaml
+++ b/.github/workflows/build-img-ghrunner-test.yaml
@@ -12,11 +12,6 @@ jobs:
     name: build-mapt
     runs-on: ubuntu-24.04
     steps:
-    - name: Prepare runner
-      shell: bash
-      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/build-img-ghrunner-test.yaml
```diff
@@ -10,13 +10,23 @@ on:
 jobs:
   build-mapt:
     name: build-mapt
-    runs-on: ubuntu-24.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, ubuntu-24.04-arm]
```
adrianriobo, line 16: As we will only test the amd64 I would build only that one? WDYT?

adrianriobo (approved):
LGTM

Issue #456, opened 2025-03-28
[bug] default value of cpu number for `azure windows create` not working somethimes


lilyLuLiu:
The Windows machine from azure, cup number is not as requested

reserve command:

```
podman run -d --name windows-create --rm \
-v ${PWD}:/workspace:z \
-e ARM_TENANT_ID=$ARM_TENANT_ID \
-e ARM_SUBSCRIPTION_ID=$ARM_SUBSCRIPTION_ID \
-e ARM_CLIENT_ID=$ARM_CLIENT_ID \
-e ARM_CLIENT_SECRET="$ARM_CLIENT_SECRET" \
-e AZURE_STORAGE_ACCOUNT="$AZURE_STORAGE_ACCOUNT" \
-e AZURE_STORAGE_KEY="$AZURE_STORAGE_KEY" \
   quay.io/redhat-developer/mapt:v0.8.2 azure \
    windows create \
     --project-name 'windows-desktop-test' \
     --backed-url azblob://crc-qenvs-state/windows-prcheck \
     --conn-details-output '/workspace' \
     --nested-virt \
     --windows-version '11' \
     --windows-featurepack '23h2-ent' \
    --tags project=openshift-local,source=github,org=crc,run=test \
    --spot \
--spot-excluded-regions southafricawest,australiacentral2
podman logs -f windows-create
```

check the cpu number of the reserved machine
`(Get-CimInstance Win32_Processor).NumberOfLogicalProcessors`

expect result:
the cpu number is 8

actual result:
sometimes the cpu number is 2

adrianriobo:
According to this https://learn.microsoft.com/en-us/azure/virtual-machines/constrained-vcpu?tabs=family-E

> To help manage your costs, Azure offers predefined VM sizes with a lowered vCPU count to help to reduce the cost of software licensing, while maintaining the same memory, storage, and I/O bandwidth.
> 
> The original size's available vCPU count can be reduced to either one half or one quarter of the original size specification. These new VM sizes have a suffix that specifies the number of available vCPUs to make them easier for you to identify
> 

From an execution hitting this issue we can see the vm size picked is `standard_e8-2as_v5` under this situation the specs for VM says `8` vCPUs (E8) but due to the reduction (2as) it got redcued to `2`.

[...]

PR #476, opened 2025-04-23
filter the vmsizes that has lowered vCPU

lilyLuLiu:
fix https://github.com/redhat-developer/mapt/issues/456

### DIFF
--- a/pkg/provider/util/instancetypes/azure.go
+++ b/pkg/provider/util/instancetypes/azure.go
@@ -174,6 +174,10 @@ func resourceSKUToVirtualMachine(res *armcompute.ResourceSKU) *virtualMachine {
 
 func filterCPUsAndMemory(cpus, memory int32, arch arch, nestedVirt bool) filterFunc {
 	return func(ctx context.Context, vm *virtualMachine, wg *sync.WaitGroup, vmCh chan<- string) {
+		// VM size is lowered with '-'suffix.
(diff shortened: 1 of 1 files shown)


pkg/provider/util/instancetypes/azure.go
```diff
@@ -185,10 +185,13 @@ func filterCPUsAndMemory(cpus, memory int32, arch arch, nestedVirt bool) filterF
 			if nestedVirt && !vm.nestedVirtSupported() {
 				return
 			}
-
 			if vm.VCPUs >= cpus && vm.Memory >= memory && vm.Arch == arch.String() &&
 				vm.baseFeaturesSupported() {
-				vmCh <- vm.Name
+				var lowerCpuPattern = `Standard.*-.*_v\d$`
```
adrianriobo, line 190: Can you add some comment on the expression to tell what is about i.e. 

```
// standard D,E and F series are the VM families
// supporting nested virtualization
var dSeriesPattern = `standardD.*v[3-6]Family$`
```

Probably it also make sense to move it from here to outside of the function, just as a var at the beginning of the file with clear information about what is this used for.

Also consider to extend the commit message to something like:

```
fix: azure wrong capacity for vm sizes. 

Here explain the problem and add the link from the issue (azure link about half capacity machines...)
```

Beyond this LGTM

adrianriobo (approved):
LGTM

PR #486, opened 2025-05-14
chore: expose param to control increase rate on spot price

### LINKED ISSUE #483: [Enhancement] Offer spot safe percentage as parameter
Currently we introduced a param to increase the history price from spot instances to allow an extra increment on bid price to ensure get the machine. That param is currently fixed at 20% but it would be great to allow control its value externally

adrianriobo:
Recently it has been observed when requesting some type of vm sizes thorugh the spot module, even when the placement score is high enough we are not getting those machines, this is mostly due to price conflicts... this commit will introduce a new parameter to set an increase rate to the spot price, so we can increase the overall spot bid

Fix #483


flacatus:
LGTM

Issue #468, opened 2025-04-15
[Enhancement] Tekton params related to owner for secret holding access information required for destroy


adrianriobo:
Owner params are used for create operation but not for destroy although as we do not set a default value for them... we are forcing users to set them on destroy... 

We can set default to empty string, but we need to validate on create operation they have value.

PR #496, opened 2025-05-29
tkn add default value for owner parameters

lilyLuLiu:
fix https://github.com/redhat-developer/mapt/issues/468


adrianriobo (approved):
LGTM

PR #503, opened 2025-06-10
snc: remove ca-cert-file flag from tekton task

anjannath:
the '--ca-cert-file' for openshift-snc create operation was removed in 362ba773a343ef5c97107f1f83391b5464edc035


adrianriobo:
On the params section there is a format about the secret holding the pull-secret...which contains info about ca-cert as optional ..can you remove that too?

Issue #473, opened 2025-04-17
[bug] Error in OpenShift CI - permission denied


zdrapela:
The last version of mapt that is working in OpenShift CI is `v0.7.4`. From `v0.8.0` on, I'm seeing `error: loading plugin list: open /opt/mapt/run/plugins: permission denied` while doing `mapt azure create`. I noticed that there might be an issue as mentioned in https://github.com/redhat-developer/mapt/pull/401#issuecomment-2682106820

See the logs for `v0.8.1`: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-redhat-developer-rhdh-main-e2e-tests-aks-helm-nightly/1903703469697536000#1:build-log.txt%3A49

Logs for `v0.9.0-dev`: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/63338/rehearse-63338-pull-ci-redhat-developer-rhdh-main-e2e-tests-aks-helm-nightly/1912790098865295360#1:build-log.txt%3A62
Test PR: https://github.com/openshift/release/pull/63338

adrianriobo:
I see... I introduced a change that I thought it may break...will check and fix it

PR #506, opened 2025-06-11
[fix] use non-root user, set group ownership and file permission

lilyLuLiu:
fix https://github.com/redhat-developer/mapt/issues/473


adrianriobo (approved):
LGTM

PR #516, opened 2025-06-19
openshift-snc: change command checking if kubeconfig is updated

### LINKED ISSUE #510: [openshift-snc] kubeconfig file is copied out too early
In #490 we added a check to verify that the `ocp-cluster-ca.service` is ran successfully and only after that we copied the `kubeconfig` since that service updates the file, but this is not working as expected and the `kubeconfig` file is copied before it has been updated

anjannath:
the ocp-cluster-ca.service after changing the cluster CA writes a new kubeconfig file with updated client certificates for the system:admin user and the token for kubeadmin user, this changes the command  that checked for the status of the ocp-cluster-ca.service to instead check if the kubeconfig file at /opt/crc/kubeconfig is created

fixes #510


adrianriobo (approved):
LGTM once tested merge, up to you

Issue #524, opened 2025-07-03
NodeHasDiskPressure error with single node cluster on AWS


fjglira:
Hi, I'm running some tests using `mapt` for the OSSM team. During the run of the test, I got some random errors during the e2e execution because the cluster shows `NodeHasDiskPressure` and `FreeDiskSpaceFailed` errors. This causes some pods to take more time to be created or not be created at all

**oc describe nodes**
```
oc describe nodes
Name:               crc
Roles:              control-plane,master,worker
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/os=linux
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=crc
                    kubernetes.io/os=linux
                    node-role.kubernetes.io/control-plane=
                    node-role.kubernetes.io/master=
                    node-role.kubernetes.io/worker=
                    node.openshift.io/os_id=rhel
                    topology.hostpath.csi/node=crc
Annotations:        csi.volume.kubernetes.io/nodeid: {"kubevirt.io.hostpath-provisioner":"crc"}
                    k8s.ovn.org/host-cidrs: ["10.0.2.53/24","192.168.126.11/24"]
                    k8s.ovn.org/l3-gateway-config:
                      {"default":{"mode":"shared","bridge-id":"br-ex","interface-id":"br-ex_crc","mac-address":"0e:f5:19:6c:58:0d","ip-addresses":["10.0.2.53/24...
                    k8s.ovn.org/node-chassis-id: 3b361556-c0a9-402d-9ed6-7d4559fb45e9
                    k8s.ovn.org/node-gateway-router-lrp-ifaddrs: {"default":{"ipv4":"100.64.0.2/16"}}
                    k8s.ovn.org/node-id: 2
                    k8s.ovn.org/node-masquerade-subnet: {"ipv4":"169.254.0.0/17","ipv6":"fd69::/112"}
                    k8s.ovn.org/node-primary-ifaddr: {"ipv4":"10.0.2.53/24"}
                    k8s.ovn.org/node-subnets: {"default":["10.217.0.0/23"]}
                    k8s.ovn.org/node-transit-switch-port-ifaddr: {"ipv4":"100.88.0.2/16"}
                    k8s.ovn.org/remote-zone-migrated: crc
                    k8s.ovn.org/zone-name: crc
                    machineconfiguration.openshift.io/controlPlaneTopology: SingleReplica
[...]

PR #525, opened 2025-07-04
Adding growpart to snc cloud cloudconfig init

fjglira:
Fixes: #524

### DIFF
--- a/pkg/provider/aws/action/openshift-snc/cloud-config
+++ b/pkg/provider/aws/action/openshift-snc/cloud-config
@@ -1,4 +1,9 @@
 #cloud-config
+bootcmd:
+  # Resize the partition (4 = /dev/nvme0n1p4 typically)
+  - growpart /dev/nvme0n1 4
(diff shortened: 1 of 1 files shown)


pkg/provider/aws/action/openshift-snc/cloud-config
```diff
@@ -16,6 +16,7 @@ runcmd:
   - chmod 0644 /opt/crc/pass_developer
   - echo "{{ .PublicIP }}" > /opt/crc/eip
   - chmod 0644 /opt/crc/eip
+  - sudo xfs_growfs -d /
```
adrianriobo, line 19: sorry, remove sudo here as this is already being executed as root during boot time

adrianriobo:
Also can you change the commit message to this format (sample said it in your words:

```
fix: openshift-snc not taking full volume size of the machine

After spin openshift-snc the mahcine was showing previos disk size, this commit will allow to adapt size to the volume attached to the machine
```

PR #526, opened 2025-07-04
feat: Return aws metadata when kind cluster is being created

flacatus:


flacatus:
@adrianriobo do u see a easy way to get the instance id from pulumi?

pkg/provider/aws/action/kind/kind.go
```diff
@@ -46,23 +46,28 @@ type kindRequest struct {
...
+
 // Create orchestrate 3 stacks:
 // If spot is enable it will run best spot option to get the best option to spin the machine
 // Then it will run the stack for windows dedicated host
-func Create(ctx *maptContext.ContextArgs, args *KindArgs) error {
-	// Create mapt Context
+func Create(ctx *maptContext.ContextArgs, args *KindArgs) (*KindResultsMetadata, error) {
 	if err := maptContext.Init(ctx, aws.Provider()); err != nil {
-		return err
+		return &KindResultsMetadata{}, err
```
adrianriobo, line 62: As returned values are *KindResultsMetadata, error...this should return nil, err

same for other returns with err

pkg/provider/aws/action/kind/kind.go
```diff
@@ -116,55 +113,106 @@ func (r *kindRequest) createHost() error {
 				awsConstants.CONFIG_AWS_NATIVE_REGION: *r.allocationData.Region}),
 		DeployFunc: r.deploy,
 	}
-	sr, _ := manager.UpStack(cs)
-	return manageResults(sr, r.prefix)
```
adrianriobo, line 120: Why did you remove this?....if beyond write those results and the price which was missing previously...just add that struct there (KindResutlsMetadata) and return it from the manageResults function no? ...all code for getting each results is already there no?

PR #528, opened 2025-07-04
feat: extra port mapping for kind cluster

psturc:
### Description
While testing a konflux running on kind cluster, we found out it'd be good to have additional ports mapped in aws so other services can be publicly accessible (e.g. tekton-results, tekton-pipelines, etc).

This change enables adding extra port mappings via new flag.

Example:
`--extra-port-mappings='[{\"containerPort\": 8080, \"hostPort\": 8080, \"protocol\": \"TCP\"}]'`

### Verification

Ran the `mapt create` command with following param:

```
--extra-port-mappings='[{"containerPort": 30012, "hostPort": 8080, "protocol": "TCP"}, {"containerPort": 30013, "hostPort": 9000, "protocol": "TCP"}]'
```

And verified that the new ports appear in the EC2 inbound rules

Screenshot 2025-07-15 at 12 07 19

and that the endpoints are accessible via exposed ports

### DIFF
--- a/cmd/mapt/cmd/aws/services/kind.go
+++ b/cmd/mapt/cmd/aws/services/kind.go
@@ -1,10 +1,13 @@
 package services
 
 import (
+	"fmt"
+
 	awsParams "github.com/redhat-developer/mapt/cmd/mapt/cmd/aws/constants"
 	"github.com/redhat-developer/mapt/cmd/mapt/cmd/params"
 	maptContext "github.com/redhat-developer/mapt/pkg/manager/context"
 	"github.com/redhat-developer/mapt/pkg/provider/aws/action/kind"
+	kindCloudConfig "github.com/redhat-developer/mapt/pkg/provider/util/cloud-config/kind"
 	"github.com/redhat-developer/mapt/pkg/util/logging"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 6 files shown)


adrianriobo:
Yeah I think the feature makes totally sense, and I checked the code quick and it seems totally fine.... just for the cmd part there is an specific type you can use to avoid all the parsing..

check this: 

https://github.com/adrianriobo/mapt/blob/ef86305923e0d9bddee44771d3e52c90cd7a8107/cmd/mapt/cmd/params/params.go#L65

https://github.com/adrianriobo/mapt/blob/ef86305923e0d9bddee44771d3e52c90cd7a8107/cmd/mapt/cmd/params/params.go#L140

https://github.com/adrianriobo/mapt/blob/fix-406/cmd/mapt/cmd/params/params.go#L127

This will simplify the code in that part

adrianriobo:
Rebase as I changed some file names and specs to facilitate spot management :pray:

adrianriobo:
Just one thing...did you try to parse the struct as json ...so we can avoid the escape chars...

something like:

```
package main

import (
	"encoding/json"
	"fmt"
	"log"

	"github.com/spf13/pflag"
	"github.com/spf13/viper"
)

type Config struct {
	Name   string   `json:"name"`
	Enabled bool     `json:"enabled"`
	Items  []string `json:"items"`
}

func main() {
	// Define a flag for the JSON config
	pflag.String("config", "", "JSON string with configuration")
	pflag.Parse()

	// Bind pflag to viper
	err := viper.BindPFlags(pflag.CommandLine)
	if err != nil {
		log.Fatalf("error binding flags: %v", err)
	}

	// Get the JSON string
	configStr := viper.GetString("config")
	if configStr == "" {
		log.Fatal("missing --config argument")
	}

	// Unmarshal JSON string into struct
	var cfg Config
	err = json.Unmarshal([]byte(configStr), &cfg)
	if err != nil {
		log.Fatalf("error parsing config JSON: %v", err)
	}

	// Use the config
	fmt.Printf("Parsed config: %+v\n", cfg)
}

**Comment by psturc** [comment]
yep, tested it like this:
```
mapt aws kind create --project-name kind-psturc-test-extra-params-1 \
--backed-url "s3://mapt-kind-bucket/mapt/kind/psturc-test-extra-params-1-$(date -u +"%s")" \
--conn-details-output /tmp/ --arch x86_64 --cpus 32 --memory 64 --version v1.32 --spot --spot-increase-rate 20 \
--timeout 20m --tags env=konflux,user=psturc \
--extra-port-mappings='[{"containerPort": 30012, "hostPort": 8080, "protocol": "TCP"}, {"containerPort": 30013, "hostPort": 9000, "protocol": "TCP"}]'
```

also added a validation of provided input values for the new flag

PR #549, opened 2025-07-29
fox: enable Load Balancer for on-demand Kind instances and fix nil pointer issues

flacatus:
…inter issues


adrianriobo:
Closed in favor of #579

PR #566, opened 2025-08-07
fix: regression for machines supporting nested virtualization on AWS

### LINKED ISSUE #563: mapt v0.9.5 `nested-virt` parameter not working
Parameter `nested-virt` parameter not working for mapt v0.9.5

Fedora from AWS :
> You need to enable virtualization in BIOS

Linux-arm64 from AWS:
>level=info msg="Setting up KVM"
level=debug msg="Trying to load kvm module"
level=error msg="Could not find cpu flags from /proc/cpuinfo"

adrianriobo:
Due to refactor on compute-request selector the value for nested virtualization was missing when filtering machines. Then even when the flag was used the type of machines returned are not taking it into consideration. This commit ensures the value is pass again in the filter. 
Fixes #563


adrianriobo:
Ok...only took me 2 days ..read this right and undestand it was about AWS ...not Azure...this is a regression b/c of the heavy refactor I did in these modules.

^^@lilyLuLiu  When the image for this PR is built can you test it? and add comment here?

psturc:
verified the latest update 👍

adrianriobo:
@lilyLuLiu @psturc thanks

PR #580, opened 2025-08-20
fix: aws public subnet extra Eip

adrianriobo:
When provisioning a public subnet it can create a NatGateway, the NatGateway requires an EIP which was created even when the NatGateway was not needed, this commit creates the Eip only if NatGateway is required


lilyLuLiu:
Tested this PR, the EIP is successfully deleted with the decommission job

lilyLuLiu (approved):

PR #595, opened 2025-09-17
fix: AWS new regions not described even when enabled

### LINKED ISSUE #593: New AWS regions not taken into account
I activated new regions in my AWS account 

- il-central-1
- ap-southeast-3

but when I request a new spot instance of type `p4de.24xlarge` or `p5en.48xlarge`, these regions are not checked.

adrianriobo:
New regions are not enable by default by AWS, those are considered opted-in regions, and even when enable they need to be explictly asked. This add them on the GetRegions funcion. 

Fix #593

### DIFF
--- a/pkg/provider/aws/data/regions.go
+++ b/pkg/provider/aws/data/regions.go
@@ -13,6 +13,7 @@ import (
 var (
 	optInStatusFilter      string = "opt-in-status"
 	optInStatusNorRequired string = "opt-in-not-required"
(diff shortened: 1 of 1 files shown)


ppitonak:
I tried image `ghcr.io/redhat-developer/mapt:pr-595` but none of these optional regions were checked
* ap-southeast-3
* eu-central-2
* eu-south-1
* eu-south-2
* il-central-1
* me-central-1

ppitonak (approved):

PR #601, opened 2025-09-24
snc: fix a panic when not using spot instance 

anjannath:


adrianriobo:
@anjannath NVM I changed some setting and it was too strcit

PR #603, opened 2025-09-24
fix: AWS PlacementeScoreData API not supported

### LINKED ISSUE #602: [Bug] PlacementScoreData API not accessible on all regions
Due to how we run parallel requests for PlacementScore we are using the same region we request data to access the API, some regions specially new opt-in ones do not offer the API so it ends up failing.

adrianriobo:
Previously we use the target region for the data as the region to access the API, but some regions do not offer the API, which is different from not having data for the PlacementScore in that region. This commit ensures all API calls to get the data are made to a fixed region where we know the API is available. Fix #602

### DIFF
--- a/pkg/provider/aws/data/spot.go
+++ b/pkg/provider/aws/data/spot.go
@@ -379,11 +379,15 @@ type placementScoreResult struct {
 	azName string
 }
 
+// Every async run will use same region to request data
+// as API is not available on all regions.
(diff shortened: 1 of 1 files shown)


ppitonak (approved):

ppitonak:
Merging is blocked
Commits must have verified signatures.

PR #607, opened 2025-09-26
snc: do not copy AMI to target region

anjannath:
cloud-importer will upload and replicate
the snc AMIs to all avilable regions, as
such when mapt is unable to find an  AMI
it is treated as an error

### DIFF
--- a/pkg/provider/aws/action/openshift-snc/openshift-snc.go
+++ b/pkg/provider/aws/action/openshift-snc/openshift-snc.go
@@ -18,7 +18,6 @@ import (
 	awsConstants "github.com/redhat-developer/mapt/pkg/provider/aws/constants"
 	"github.com/redhat-developer/mapt/pkg/provider/aws/data"
 	"github.com/redhat-developer/mapt/pkg/provider/aws/modules/allocation"
-	amiCopy "github.com/redhat-developer/mapt/pkg/provider/aws/modules/ami"
 	"github.com/redhat-developer/mapt/pkg/provider/aws/modules/ec2/compute"
 	"github.com/redhat-developer/mapt/pkg/provider/aws/modules/iam"
 	"github.com/redhat-developer/mapt/pkg/provider/aws/modules/network"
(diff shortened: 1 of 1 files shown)


pkg/provider/aws/action/openshift-snc/openshift-snc.go
```diff
@@ -106,10 +105,9 @@ func Create(mCtxArgs *mc.ContextArgs, args *OpenshiftSNCArgs) (_ *OpenshiftSncRe
 	if err != nil {
 		return nil, err
 	}
-	// Manage AMI offering / replication
+	// check if AMI exists
 	amiName := amiName(&args.Version, &args.Arch)
-	if err = manageAMIReplication(mCtx, &args.Prefix,
-		&amiName, r.allocationData.Region, &args.Arch); err != nil {
+	if err = checkAMIExists(&amiName, r.allocationData.Region, &args.Arch); err != nil {
```
adrianriobo, line 110: mmm may it would be interesting get the AMI info, so we can set it on the request struct and do not need to search for it again on the deploy func?

anjannath: so to get the AMI (`ec2.LookupAMIResult`) we need the `pulumi.Context` which is only available in the deploy func

the `checkAMIExists` function is using the aws SDK and not using pulumi aws provider functions, so it doesn't need the `pulumi.Context`

adrianriobo: mm but what do we need? AMI ID or ARN...you can use it as regultar string in the request and then use pulumi.String(string) no?

adrianriobo (approved):

pkg/provider/aws/action/openshift-snc/openshift-snc.go
anjannath: we need the type `ec2.LookupAMIResult` in the [`ComputeRequest`](https://github.com/redhat-developer/mapt/blob/6c86268697271a65e91482cc244dd3fde309c680/pkg/provider/aws/modules/ec2/compute/compute.go#L50) struct

currently in the `deploy` func there's call to [`GetAMIbyName`](https://github.com/redhat-developer/mapt/blob/6c86268697271a65e91482cc244dd3fde309c680/pkg/provider/aws/action/openshift-snc/openshift-snc.go#L173) and this returns the `ec2.LookupAMIResult` which is then set on the `computeRequest` struct

PR #613, opened 2025-10-01
fix: unify quotes to avoid bash syntax errors with empty strings

ppitonak:
* params with default value of empty string are not evaluated correctly

It used to render script like this
```
if [[ create == "create" ]]; then
    cmd+="--conn-details-output /opt/host-info "
    if [[  != "" ]]; then         # notice left-hand operand missing, bash syntax error
      cmd+="--compute-sizes  "    # empty-string value not rendered which breaks mapt
    else
```


adrianriobo (approved):
LGTM

PR #621, opened 2025-10-07
azure: implement DefaultHostiPlace method

anjannath:
adds implementation for the missing method to
satisfy the provider interface


adrianriobo (approved):
LGTM

PR #632, opened 2025-10-08
Mask AWS and Azure credentials in debug mode

ppitonak:
Previously when debug mode was active, AWS credentials were displayed which is very insecure because Tekton logs could be archived for long time.

Now only the first and last character of sensitive credentials are displayed.


adrianriobo:
Would you mind to add this to azure tasks as well....if not this is something we will face in the future too ...

ppitonak:
I pushed the updates also for Azure

PR #637, opened 2025-10-16
tkn: update task for azure windows desktop

anjannath:
replace the use of flag `--vmsize` with `--compute-sizes`
flag name was updated in bd2d62640c418a3a25ae0eccd0f1be21d23ef717

### DIFF
--- a/tkn/infra-azure-windows-desktop.yaml
+++ b/tkn/infra-azure-windows-desktop.yaml
@@ -90,8 +90,8 @@ spec:
     - name: nested-virt
       description: Use cloud instance that has nested virtualization support
       default: 'false'
-    - name: vmsize
-      description: size for the machine
+    - name: compute-sizes
+      description: Comma separated list of sizes for the machine
       default: "''"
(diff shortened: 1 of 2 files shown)


albfan:
In any case, this should use `-n` (not empty string) instead of `-z` (empty string) for:

    if [[ -z $(params.vmsize) ]]; then cmd+="--vmsize $(params.vmsize) "; fi

adrianriobo:
@albfan is this breaking your pipelines????

albfan:
I'm not sure what happens, probably related with fixing empty strings:

https://github.com/redhat-developer/mapt/pull/613

In any case, problems with git resolvers on our openshift cluster, make us apply all this task, so we modify manually until we resolve network problems

adrianriobo:
@ppitonak https://github.com/redhat-developer/mapt/pull/613#issuecomment-3356024775 do you recall what test did you do,...I mean did you test both....compute-size based and non compute-size based?

PR #661, opened 2025-11-14
explicitly set workingDir in step host-info-secret

ppitonak:
in some cluster configurations, default path could cause "permission denied"


adrianriobo (approved):
LGTM

PR #666, opened 2025-11-19
fix: Allow any len for compute sizes param when asking for spot

### LINKED ISSUE #622: [bug] Too many required spot type machines when setting compute-sizes
Recently we included a number of spot types to ensure we will get machines.

The number https://github.com/redhat-developer/mapt/blob/de738943390756fffcad11682668fccc562817d5/pkg/provider/aws/data/spot.go#L57 is fixed to 8 which will works when asking for machines based on params (CPU, Memory,...)

But when using specific compute-sizes if the number of machines is less than 8 ...it is failing.

We need to check if compute-sizes is set, and if that is the case dynamically adjust the field

adrianriobo:
Previously there was a fixed value to 8 on spot calculations, now if computeSizes are pass as param and the number of sizes is less than 8 it will use the len of compute sizes. This means now we will be able to use any number of compute sizes. 
Fixes #622

Signed-off-by: Adrian Riobo <ariobolo@redhat.com>

### DIFF
--- a/pkg/provider/aws/data/spot.go
+++ b/pkg/provider/aws/data/spot.go
@@ -54,7 +54,7 @@ var placementScores = []placementScoreSpec{
 // While calculating the spot price and the types of machines
 // to be requested we use the first n number of types per Az
 // to control how many types will be used this var is used
-var numberOfTypesForSpot = 8
+var maxNumberOfTypesForSpot = 8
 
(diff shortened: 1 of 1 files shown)


adrianriobo:
Yeah actually It is not imposed, I just tried with 2 and it worked..so let me change the commit message

Issue #630, opened 2025-10-08
[Bug] Interrupting `mapt` doesn't cancel the underlying `pulumi` task


kpouget:
I'm not exactly sure what `cancel the `pulumi` task` means, but...

Whenever I need to interrupt `mapt` because the `main-ssh-readiness-aos-cmd` or `main-ssh-readiness-aos-cmd` commands don't succeed (because my image is broken), I have to run the following `pulumi` command:
```
cd stackOpenshiftSNC-$NAME; 
pulumi cancel  /stackOpenshiftSNC-$NAME --yes
```

`mapt`  should do this automatically, so that I can launch the `destroy` command immediately.

adrianriobo:
I think this one if fixed too by #675 as now if you ctrl+c the execution pullumi finalize gracefully as so no lock is left. Based on info from #675.

I am gonna close this one, if you still see the behavior with latest image can you re open it?

PR #675, opened 2025-12-02
Add full context propagation and proper interrupt handling

jangel97:
This MR adds full context propagation to MAPT so that interrupts (**SIGINT/SIGTERM**, e.g., Ctrl+C) cleanly cancel ongoing Pulumi tasks instead of leaving them running in the background. This completes the fix for **#630**.

### **Changes**

* CLI now creates a cancellable root context.
* Added unified signal handling for **SIGINT/SIGTERM**.
* All manager stack functions now require `ctx`.
* All AWS/Azure provider operations updated to forward that context.
* Replaced all remaining `context.Background()` usages.

### **Why**

Previously, interrupting MAPT didn’t cancel the underlying Pulumi task, often leaving partial resources and inconsistent state. Cancellation is now fully propagated to Pulumi, ensuring immediate, deterministic shutdown.

### **Testing**

* Verified Ctrl+C cancels `aws fedora create` mid-update.
  Command used:

  ```bash
  out/mapt aws fedora create \
    --project-name aipcc-6664-2 \
    --backed-url file:///$PWD/state \
    --conn-details-output $PWD/outputs \
    --compute-sizes g5.xlarge,g5.2xlarge,g4dn.xlarge \
    --cpus 1 \
    --memory 8 \
    --spot-excluded-regions us-east-1a,us-east-1b,us-east-1c,us-east-1d,us-east-1e,us-east-1f,us-east-2a,us-east-2b,us-east-2c
  ```
* Pulumi logs reflect clean cancellation events.
* Destroy operations behave correctly after interruption.

### **Issue**

Fixes **#630**.

### DIFF
--- a/cmd/mapt/cmd/aws/hosts/fedora.go
+++ b/cmd/mapt/cmd/aws/hosts/fedora.go
@@ -48,6 +48,7 @@ func getFedoraCreate() *cobra.Command {
 			}
 			return fedora.Create(
 				&maptContext.ContextArgs{
+					Context:       cmd.Context(),
 					ProjectName:   viper.GetString(params.ProjectName),
 					BackedURL:     viper.GetString(params.BackedURL),
 					ResultsOutput: viper.GetString(params.ConnectionDetailsOutput),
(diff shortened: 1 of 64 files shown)


cmd/mapt/cmd/aws/hosts/fedora.go
```diff
@@ -46,7 +47,8 @@ func getFedoraCreate() *cobra.Command {
 			if err := viper.BindPFlags(cmd.Flags()); err != nil {
 				return err
 			}
-			return fedora.Create(
+			mCtx, err := maptContext.Init(
```
adrianriobo: :thinking: but why do you need to pass this as param?  ...can you init the context inside the context.Init function?

adrianriobo:
Hey got one question did you try this out, I mean what is status for state file? if ctrl+c if you try to destroy them afterwards is it locked?

PR #681, opened 2025-12-04
feat: IBM Cloud support

### LINKED ISSUE #471: [Feature] Support for IBM VPC instances
I would be great if we could support create VPC instances on IBM cloud.

adrianriobo:
Fix #471


docs/ibmcloud/ibm-power.md
```diff
@@ -0,0 +1,55 @@
+# Overview
+
+This actions will handle provision Ubuntu ppc64 machines on ibm cloud power vs.
```
ppitonak: What is `vs`?

ppitonak (commented):
What's the reason for not following established command conventions? For AWS we have 
```
mapt aws rhel create --arch=...
```

But for IBM we have
```
mapt ibmcloud ibm-power create 
mapt ibmcloud ibm-z create
```

Is there a technical reason why it cannot be implemented similarly to AWS and treat the cloud-specific differences as implementations details?

docs/ibmcloud/ibm-power.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Outputs
+
+* It will crete an instance and will give as result several files located at path defined by `--conn-details-output`:
+
+  * **host**: host for the Windows machine (lb if spot)
+  * **username**: username to connect to the machine
+  * **id_rsa**: private key to connect to machine
+
+* Also, it will create a state folder holding the state for the created resources at azure, the path for this folder is defined within `--backed-url`, the content from that folder it is required with the same project name (`--project-name`) in order to destroy the resources.
```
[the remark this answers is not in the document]
adrianriobo: I just realized yesterday IBM has a storage containter service which is S3 interface compatible so we can use the s3 backed url and the IBM storage (s3 alike). 

The thing is to setup access the ENVs pulumi expects are AWS_... which could mislead .. I am thinking to create our own S3_...envs... for the backed ...then for aws if S3 envs will be set those will take precedence over AWS_... and for IBM those will be the ones use to set the access WDYT?

pkg/provider/ibmcloud/data/vpcimages.go
```diff
@@ -0,0 +1,71 @@
...
+	"strings"
+
+	"github.com/IBM/vpc-go-sdk/vpcv1"
+
+	"github.com/IBM/go-sdk-core/v5/core"
+)
+
+const (
+	VPC_ARCH_X86_64 vpcArch = "amd64"
+	VPC_ARCH_IBMZ   vpcArch = "s390x"
```
srija-ganguly, line 16: drift of keywords from the files constants.go, should the constants package be imported here instead? or should the constants package be completely removed because they are not used anywhere?

pkg/provider/ibmcloud/services/power/power.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func waitForInstance(pc *v.IBMPIInstanceClient, instanceId string) {
+	for i := 0; i < 30; i++ { // retry up to ~5 minutes
+		inst, err := pc.Get(instanceId)
+		if err == nil && inst.Health.Status == "WARNING" {
+			fmt.Println("Instance ready")
+			break
+		}
+		fmt.Println("Instance not ready, retrying in 10s...")
```
srija-ganguly: waitForInstance silently succeeds even on failure. If the instance never reaches "WARNING" health status, no error is returned, and the caller proceeds as if everything is fine. Should it return an error on timeout?

pkg/provider/ibmcloud/services/power/power.go
```diff
@@ -0,0 +1,91 @@
...
+		UserAccount: os.Getenv("IBMCLOUD_ACCOUNT"),
+		Zone:        os.Getenv("IC_ZONE"),
+		URL:         powerURL(os.Getenv("IC_REGION")),
+		Debug:       mCtx.Debug(),
+	}
+	session, err := ps.NewIBMPISession(options)
+	if err != nil {
+		return nil, err
+	}
+	return v.NewIBMPIInstanceClient(context.Background(), session, cloudInstanceId), nil
```
srija-ganguly: do we want the user to have the ability to cancel these calls (ctrl+c propagation) by using mCtx.Context() instead?

pkg/provider/ibmcloud/action/ibm-power/cloud-config
```diff
@@ -0,0 +1,81 @@
...
+        resource:
+          attributes:
+            - key: appcode
+              value: "{{.AppCode}}"
+              action: upsert
+            - key: host.name
+              value: "${env:HOSTNAME}"
+              action: upsert
+            - key: com.redhat.otel.auth_token
+              value: "{{.OtelAuthToken}}"
```
srija-ganguly: security concern: this is rendered directly into the YAML config and passed as base64-encoded user data which is accessible on the IBM Cloud by any process on the instance. Same pattern in ibm-z/cloud-config

.golangci.yml
```diff
@@ -1,8 +1,23 @@
 version: "2"
 
 run:
-  timeout: 10m
+  timeout: 20m
+  skip-dirs:
+    - vendor
+    - tools/vendor
+  skip-files:
+    - pkg/provider/ibmcloud/action/powervs/powervs.go
```
srija-ganguly: is this path still valid?

srija-ganguly:
The security group + inbound/outbound rules + floating IP setup is repeated three times — in network/network.go, ibm-power/ibm-power.go:deployBastion(), and ibm-z/ibm-z.go:deployWithExistingSubnet() — and could be extracted to avoid duplications, but the functions themselves serve different purposes.

pkg/provider/ibmcloud/data/vpcimages.go
adrianriobo: done

pkg/provider/ibmcloud/services/power/power.go
adrianriobo: done

pkg/provider/ibmcloud/action/ibm-power/cloud-config
adrianriobo: done

.golangci.yml
adrianriobo: removed

deekay2310:
Tested the following workflows on VMs provisioned through this change:
  
  - s390x (Ubuntu 22.04): Built the GitHub Actions runner from source, registered it as a self-hosted runner, and successfully ran workflows. Runner installed at /opt/runner-cache, configured and started with config.sh + svc.sh.
  - ppc64le (RHEL): Same flow — built runner from source via configure-runner.sh, registered and ran workflows successfully.

docs/ibmcloud/ibm-power.md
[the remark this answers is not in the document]
adrianriobo: this will be implemented in a follow up

docs/ibmcloud/ibm-power.md
```diff
@@ -0,0 +1,55 @@
...
+### Container
+
+```bash
+podman run -d --name ibm-power \
+        -v ${PWD}:/workspace:z \
+        -e IBMCLOUD_ACCOUNT=XXX \
+        -e IBMCLOUD_API_KEY=XXX \
+        -e IC_REGION=us-south \
+        -e IC_ZONE=dal12 \ # power vs use this notation
+        quay.io/redhat-developer/mapt:1.0.0 ibm-power create \
```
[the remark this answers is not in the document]
adrianriobo: fixed

pkg/provider/ibmcloud/action/ibm-z/ibm-z.go
```diff
@@ -0,0 +1,405 @@
...
+		Keys:          pulumi.StringArray{pik.ID()},
+		PrimaryNetworkInterface: &ibmcloud.IsInstancePrimaryNetworkInterfaceArgs{
+			Subnet: n.Subnet.ID(),
+			SecurityGroups: pulumi.StringArray{
+				n.SecurityGroup.ID(),
+			},
+		},
+	}
+	if r.otelAppCode != "" && r.otelAuthToken != "" {
+		ud, err := izUserData(r.otelAppCode, r.otelAuthToken, r.otelEndpoint, r.otelIndex, r.otelExtraAttrs)
```
jrusz, line 195: **Otel guard mismatch — silent no-op when `--otel-index` is omitted**

The Go guard here checks `otelAppCode != "" && otelAuthToken != ""`, but the cloud-config template gates all otel content on `{{if and .AppCode .OtelAuthToken .OtelIndex}}` — requiring `OtelIndex` too.

If a user passes `--otel-app-code` and `--otel-auth-token` but forgets `--otel-index`, this guard passes and the template is rendered, but the template conditional evaluates to false (empty string is falsy). The result is a MIME-wrapped cloud-config with no otel setup and no feedback to the user that it was skipped.

Same pattern appears at line 282 (existing-subnet path) and in `ibm-power/ibm-power.go:163` (though ibm-power is slightly less affected since the route setup in runcmd still runs regardless).

Suggestion: either add `OtelIndex` to this Go guard, give `--otel-index` a default value (like `--otel-endpoint` has), or log a warning when AppCode + AuthToken are set but Index is empty.

pkg/provider/ibmcloud/services/power/power.go
```diff
@@ -0,0 +1,98 @@
+package power
```
jrusz, line 1: This package doesn't appear to be imported anywhere in the codebase — the actual PowerVS provisioning is done through Pulumi in `action/ibm-power/ibm-power.go`. It also duplicates `powerURL` (defined identically in `data/piimages.go:16-18`).

If this was scaffolding for a non-Pulumi path, might be worth either wiring it up or removing it to avoid maintenance confusion.

jrusz (commented):
Reviewed the non-vendor changes (~2,250 lines across 24 files). The architecture is solid — good extraction of shared network helpers, proper cloud-init MIME wrapping for IBM Z, and thorough otel collector configuration. Leaving a few observations inline for your consideration.

Note: I'm aware of the existing review threads from @ppitonak, @srija-ganguly, and @deekay2310 — I didn't duplicate any of those.

pkg/provider/ibmcloud/action/ibm-z/ibm-z.go
```diff
@@ -0,0 +1,405 @@
...
+				func(pni ibmcloud.IsInstancePrimaryNetworkInterface) string {
+					return *pni.Id
+				},
+			).(pulumi.StringOutput),
+		})
+	if err != nil {
+		return err
+	}
+	ctx.Export(fmt.Sprintf("%s-%s", *r.prefix, outputHost), n.Floatingip.Address)
+	return nil
```
jrusz, line 224: Minor: `deployWithExistingSubnet` (line 312) includes a `remote.NewCommand` readiness ping that waits for SSH availability before reporting success. This `deploy` path (auto-provision VPC) does not, even though it also assigns a floating IP. Other providers (AWS RHEL, Windows, Mac) consistently include readiness checks. Worth adding for consistency — without it, this path may report "done" before the instance is actually SSH-reachable.

pkg/provider/ibmcloud/ibmcloud.go
```diff
@@ -0,0 +1,88 @@
...
+	return nil, fmt.Errorf("missing default value for IBM Cloud Region: IC_REGION")
+}
+
+func (a *IBMCloud) Zone() (*string, error) {
+	hp := os.Getenv("IC_ZONE")
+	if len(hp) > 0 {
+		return &hp, nil
+	}
+	return nil, fmt.Errorf("missing default value for IBM Cloud Region: IC_ZONE")
+}
```
jrusz, line 37: Nit: error message says "IBM Cloud Region" but should say "IBM Cloud Zone" (the `IC_REGION` check on line 28 correctly says Region).
```suggestion
	return nil, fmt.Errorf("missing default value for IBM Cloud Zone: IC_ZONE")
```

docs/ibmcloud/ibm-power.md
```diff
@@ -0,0 +1,147 @@
...
+      --ghactions-runner-repo string         Full URL of the repository where the Github Actions Runner should be registered
+      --ghactions-runner-token string        Token needed for registering the Github Actions Runner token
+  -h, --help                                 help for create
+      --it-cirrus-pw-labels stringToString   additional labels to use on the persistent worker (--it-cirrus-pw-labels key1=value1,key2=value2) (default [])
+      --it-cirrus-pw-token string            Add mapt target as a cirrus persistent worker. The value will hold a valid token to be used by cirrus cli to join the project.
+      --otel-app-code string                 OpenTelemetry appcode identifier (e.g. MAPT-001); when set together with --otel-auth-token, installs the otelcol-contrib filelog collector on the instance
+      --otel-auth-token string               OpenTelemetry authentication token (UUID) used to authenticate against the OTLP endpoint
+      --otel-endpoint string                 OTLP HTTP endpoint to export logs to (default "https://otel-input.corp.redhat.com")
```
jrusz, line 47: The help output here is missing `--otel-index` and `--otel-extra-attrs` flags, which are registered in the CLI code (`hosts/ibm-power.go:78-79`). Same omission in `docs/ibmcloud/ibm-z.md`. Since the otel config silently requires `--otel-index` (see other comment), documenting it here would help users avoid misconfiguration.

deekay2310 (approved):
infra provisioning works for me

PR #774, opened 2026-04-17
fix(tkn): param accelerator was ignored with compute-sizes specified

ppitonak:
For spot machines with compute sizes specified, mapt chose CUDA image even though accelerator was set to ROCm

Tested with both CUDA and ROCm


adrianriobo (approved):
LGTM

PR #778, opened 2026-04-20
feat(release): automate binary release on tag push

rishupk:
The release process was manual before. This automates it on tag push.

Each matrix runner now extracts the mapt binary from the built container image using podman cp and uploads it as mapt-linux-amd64 or mapt-linux-arm64. A new release job picks those up, generates a sha256 checksums file, and creates a GitHub Release with everything attached.

The flattened OCI tarball artifacts are gone. Instead, each runner pushes its own arch image directly, and the manifest job stitches them together. Three new Makefile targets support this: oci-push-amd64, oci-push-arm64, and oci-push-manifest.

### DIFF
--- a/.github/workflows/build-oci.yaml
+++ b/.github/workflows/build-oci.yaml
@@ -53,6 +53,16 @@ jobs:
           else
             echo "ARCH_TYPE=amd64" >> $GITHUB_ENV
           fi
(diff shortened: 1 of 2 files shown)


adrianriobo (approved):
LGTM

PR #787, opened 2026-04-28
fix(release): add --repo flag to gh release create

rishupk:
The `release` job was failing with `fatal: not a git repository` because `gh release create` needs git context to determine the target repo. The job only downloads binary artifacts — it never checks out the code. Adding `--repo "$GITHUB_REPOSITORY"` provides that context via the GitHub API without requiring a full checkout.

Fixes: https://github.com/redhat-developer/mapt/actions/runs/25001661831/job/73215210704


adrianriobo (approved):
LGTM

PR #796, opened 2026-05-12
fix(release): use --notes-start-tag for correct changelogs

rishupk:
`--generate-notes` falls back to v0.6.0 because each release has a `chore(cut)` commit that never gets merged back to main, so no recent release tag is an ancestor of the current tag. This fetches the previous release tag explicitly and passes it via `--notes-start-tag`.

Fixes: [AIPCC-16038](https://redhat.atlassian.net/browse/AIPCC-16038)


adrianriobo (approved):
LGTM

Issue #695, opened 2025-12-16
[bug] The secret files generated by `mapt` don't end with a EOL


kpouget:
The files created by `mapt` from the pulumi secrets don't end by a new line:
```
adminusername  adminuserpassword  host username  userpassword
```

As a best practice, the `\n` should be appended at the end of the file.

see the answers of https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

PR #799, opened 2026-05-13
fix: remove trailing newline from connection detail output files

jangel97:
## Summary

The `writeOutput` function in `pkg/provider/util/output/output.go` appends `"\n"` to every value written to connection detail files (`host`, `username`, etc.). This trailing newline propagates through the Tekton task pipeline and causes SSH connection failures in downstream consumers.

## Root cause

The full chain:

1. **mapt** writes `fedora\n` to `/opt/host-info/username` (the `+"\n"` on [line 25](https://github.com/redhat-developer/mapt/blob/main/pkg/provider/util/output/output.go#L25))
2. The **Tekton `host-info-secret` step** base64-encodes the file contents into a Kubernetes Secret — the newline is faithfully encoded (`ZmVkb3JhCg==` instead of `ZmVkb3Jh`)
3. **Kubernetes** decodes the Secret value back via `secretKeyRef` and sets the env var with the newline preserved: `USERNAME=fedora\n`
4. Downstream tasks build an SSH command: `ssh_cmd="ssh ... ${USERNAME}@${HOST}"`  
   With the newline, this becomes: `ssh ... fedora\n@ec2-host\n`
5. When `${ssh_cmd}` is expanded **unquoted**, bash word-splits on the newline, producing:
   ```
   [ssh] [...] [fedora] [@ec2-host]
   ```
6. SSH interprets `fedora` as the hostname → **`"Could not resolve hostname fedora: Name or service not known"`**

## Reproduction

Verified inside the `quay.io/redhat-developer/mapt:v1.0.0-dev` container:

```
=== WITH newline ===
ssh_cmd with cat -v:
ssh -o StrictHostKeyChecking=no fedora
@ec2-1-2-3-4.compute.amazonaws.com
Word split:
  [ssh]  [-o]  [StrictHostKeyChecking=no]  [fedora]  [@ec2-1-2-3-4.compute.amazonaws.com]

=== WITHOUT newline ===
ssh_cmd with cat -v:
ssh -o StrictHostKeyChecking=no fedora@ec2-1-2-3-4.compute.amazonaws.com
Word split:
  [ssh]  [-o]  [StrictHostKeyChecking=no]  [fedora@ec2-1-2-3-4.compute.amazonaws.com]
```

## Fix

Remove the `+"\n"` from `writeOutput`. The output files are only consumed by the Tekton `host-info-secret` step which pipes them through `base64 -w0` — no consumer depends on a trailing newline.

### DIFF
--- a/tkn/template/infra-aws-mac.yaml
+++ b/tkn/template/infra-aws-mac.yaml
@@ -294,15 +294,15 @@ spec:
         cat <<EOF >> host-info.yaml
         type: Opaque
         data:
-          host: $(cat /opt/host-info/host | base64 -w0)
-          username: $(cat /opt/host-info/username | base64 -w0)
-          id_rsa: $(cat /opt/host-info/id_rsa | base64 -w0)
+          host: $(cat /opt/host-info/host | tr -d '\n\r' | base64 -w0)
+          username: $(cat /opt/host-info/username | tr -d '\n\r' | base64 -w0)
+          id_rsa: $(cat /opt/host-info/id_rsa | tr -d '\n\r' | base64 -w0)
         EOF
         if [[ $(params.airgap) == "true" ]]; then
         cat <<EOF >> host-info.yaml
-          bastion-host: $(cat /opt/host-info/bastion_host | base64 -w0)
-          bastion-username: $(cat /opt/host-info/bastion_username | base64 -w0)
-          bastion-id_rsa: $(cat /opt/host-info/bastion_id_rsa | base64 -w0)
+          bastion-host: $(cat /opt/host-info/bastion_host | tr -d '\n\r' | base64 -w0)
+          bastion-username: $(cat /opt/host-info/bastion_username | tr -d '\n\r' | base64 -w0)
+          bastion-id_rsa: $(cat /opt/host-info/bastion_id_rsa | tr -d '\n\r' | base64 -w0)
         EOF
         fi
 
--- a/tkn/infra-aws-eks.yaml
+++ b/tkn/infra-aws-eks.yaml
@@ -289,7 +289,7 @@ spec:
         cat <<EOF >> cluster-info.yaml
         type: Opaque
         data:
-          kubeconfig: $(cat /opt/cluster-info/kubeconfig | base64 -w0)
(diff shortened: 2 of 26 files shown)


adrianriobo:
@anjannath ^^ why did you added the newline? 

 Also @jangel97 that was added in the last commit, so instead of new commit can you just revert?

adrianriobo:
Ok, so I think even better do not revert but add sanitizers within the tekton tasks i.e. https://github.com/redhat-developer/mapt/blob/9fc64db6667402c4f3bddc5fe639f0c375db6b4e/tkn/template/infra-aws-fedora.yaml#L347 

trim the value before encoding so we will ensure it will work

anjannath:
> @anjannath ^^ why did you added the newline?

this was added to resolve issue #695 there it has reasoning around why new-line should be appended to end-of-file
> A [text file](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_392), under unix, consists of a series of [lines](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_205), each of which ends with a [newline character](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_238) (\n). A file that is not empty and does not end with a newline is therefore not a text file.

tkn/template/infra-aws-mac.yaml
```diff
@@ -294,14 +294,14 @@ spec:
         cat <<EOF >> host-info.yaml
         type: Opaque
         data:
-          host: $(cat /opt/host-info/host | base64 -w0)
-          username: $(cat /opt/host-info/username | base64 -w0)
```
adrianriobo, line 298: It does not affect host?, that file should have now a new line too

jangel97: Yes, host is also affected `output.go appends '\n' to all values written (host, username, id_rsa, etc.), so both need the trim. wdyt?

adrianriobo: yeah basically you wanna ensure any base64 encoding is trim..so basically any line that | base64 needs to be trimmed

adrianriobo (approved):
LGTM

PR #800, opened 2026-05-14
snc: update the AMI regex to remove `-` after arch

anjannath:
in the first region where the AMI is uploaded the AMI name don't contain the region suffix and no separator

in all the other regions where the AMI is copied from the first region it contains the region as suffix and a `-` added before it to the AMI name, the new  regex will handle both the variants


adrianriobo (approved):
LGTM

PR #801, opened 2026-05-14
Revert "fix: sanitize host and username values before base64 encoding in secrets"

jangel97:
## Summary

- Reverts #799 — the `tr -d '\n\r'` was applied to **all** `base64 -w0` lines including `id_rsa`, which strips internal newlines from PEM private keys and breaks SSH authentication (`Load key "id_rsa": error in libcrypto`)
- Will follow up with a corrected fix that only trims single-line values (host, username, passwords) and leaves multi-line values (id_rsa, kubeconfig) untouched

### DIFF
--- a/tkn/infra-aws-eks.yaml
+++ b/tkn/infra-aws-eks.yaml
@@ -289,7 +289,7 @@ spec:
         cat <<EOF >> cluster-info.yaml
         type: Opaque
         data:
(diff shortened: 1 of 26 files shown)


adrianriobo (approved):
LGTM

adrianriobo (approved):
LGTM

PR #806, opened 2026-05-19
feat(profile): add --operator-channel and --catalog-source flags

jangel97:
## Summary
- Add `--operator-channel` flag to override the OLM subscription channel per operator (e.g. `--operator-channel serverless-operator=candidate`)
- Add `--catalog-source` flag to use a custom index image per operator (e.g. `--catalog-source nfd=quay.io/my-team/index:test`), creating a CatalogSource CR automatically
- Validation runs before infrastructure provisioning to fail fast on invalid inputs
- Override logic is centralized in `installOperator()` — no profile files need changes
- Docs updated with usage examples and operator package name reference table

## Test plan
- [x] Build passes
- [x] Validation catches empty keys/values before provisioning
- [x] End-to-end test with `--operator-channel` override
- [x] End-to-end test with `--catalog-source` override (NFD installed from custom index image hosted on ttl.sh, verified CatalogSource pod running and subscription pointing to custom catalog)

### DIFF
--- a/cmd/mapt/cmd/aws/services/snc.go
+++ b/cmd/mapt/cmd/aws/services/snc.go
@@ -27,6 +27,11 @@ const (
 
 	sncProfile     = "profile"
 	sncProfileDesc = "comma separated list of profiles to apply on the SNC cluster. Profiles available: virtualization, serverless-serving, serverless-eventing, serverless, servicemesh, ai, nvidia. The ai profile automatically includes servicemesh and serverless-serving as prerequisites and raises the minimum instance size to 16 vCPUs. The nvidia profile installs NFD and the NVIDIA GPU Operator"
+
+	operatorChannel     = "operator-channel"
+	operatorChannelDesc = "override the OLM subscription channel for an operator (--operator-channel serverless-operator=preview,nfd=4.17)"
+	catalogSource       = "catalog-source"
+	catalogSourceDesc   = "override the OLM catalog source with a custom index image (--catalog-source serverless-operator=quay.io/my-org/my-index:latest)"
 )
 
 func GetOpenshiftSNCCmd() *cobra.Command {
@@ -92,7 +97,9 @@ func createSNC() *cobra.Command {
 					PullSecretFile:          viper.GetString(pullSecretFile),
 					Timeout:                 viper.GetString(params.Timeout),
 					ServiceEndpoints:        params.NetworkServiceEndpoints(),
-					Profiles:                profiles}); err != nil {
+					Profiles:         profiles,
+					OperatorChannels: viper.GetStringMapString(operatorChannel),
+					CatalogSources:   viper.GetStringMapString(catalogSource)}); err != nil {
 				return err
 			}
 			return nil
@@ -107,6 +114,8 @@ func createSNC() *cobra.Command {
 	flagSet.StringP(params.Timeout, "", "", params.TimeoutDesc)
 	flagSet.StringToStringP(params.Tags, "", nil, params.TagsDesc)
 	flagSet.StringSliceP(sncProfile, "", []string{}, sncProfileDesc)
+	flagSet.StringToStringP(operatorChannel, "", nil, operatorChannelDesc)
+	flagSet.StringToStringP(catalogSource, "", nil, catalogSourceDesc)
 	params.AddComputeRequestFlags(flagSet)
 	params.AddSpotFlags(flagSet)
 	params.AddNetworkFlags(flagSet, awsParams.ServiceEndpointsDesc)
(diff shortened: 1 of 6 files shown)


adrianriobo:
## Code Review — `--operator-channel` and `--catalog-source` flags                                                                                                                                        
                                                            
  ### Overview                                                                                                                                                                                              
  Adds two CLI flags to the SNC provisioner that let callers override OLM subscription channels and inject custom CatalogSource index images per operator. This is the mechanism needed to install OpenShift
   AI from a custom FBC bundle — you'd use `--catalog-source rhods-operator=quay.io/your-org/fbc-index:tag` combined with `--operator-channel rhods-operator=fast`.                                         
                                                            
  ---                                                                                                                                                                                                       
                                                            
  ### Design — Good

  - **Centralized override in `installOperator`** — overrides are applied in one place, no per-operator special-casing needed.                                                                              
  - **CatalogSource naming** with `mapt-cs-<pkg>-<sha256[:8]>` is deterministic and avoids collisions across re-runs (Pulumi can diff it).
  - **`DeletedWith` propagation** — CatalogSource CRs are also marked with the compute resource, so they're skipped on destroy.                                                                             
  - **`ValidateOperatorOverrides` runs before infra provisioning** — fail-fast without spending money on an EC2 instance.                                                                                   
  - **Incompatibility documentation** (`ProfileOpenShiftAI` + `ProfileServiceMesh`) is well commented.                                                                                                      
                                                                                                                                                                                                            
  ---                                                                                                                                                                                                       
                                                                                                                                                                                                            
  ### Issues                                                

  **1. No CatalogSource readiness wait — likely to break RHOAI installs (critical)**                                                                                                                        
   
  `ensureCatalogSources` creates the CatalogSource CR and returns. OLM then needs to pull the FBC image and spin up the gRPC server (30–90 seconds). The Subscription is created with                       
  `pulumi.DependsOn(cs.Resource)` which only waits for the K8s object to exist, not for OLM to index it. The subscription will fail with "no operators found" until the catalog is ready.
                                                                                                                                                                                                            
  Suggested fix: reuse the existing `waitForCRCondition` pattern to poll `.status.connectionState.lastObservedState == "READY"` before returning from `ensureCatalogSources`.                               
   
  **2. `MaxCPUs` logic is inverted for multi-profile use**                                                                                                                                                  
                                                            
  The function returns the upper CPU cap, but when two profiles each set `maxCPUs`, the effective cap should be `min` (you must satisfy both constraints), not `max`. Only `ProfileNvidia` sets `maxCPUs`   
  today so it doesn't bite yet, but the logic is wrong:     
                                                                                                                                                                                                            
  ```go                                                     
  // Should return the lowest non-zero cap, not the highest
  func MaxCPUs(profiles []string) int32 {                                                                                                                                                                   
      var cap int32                                                                                                                                                                                         
      for _, p := range profiles {                                                                                                                                                                          
          if effect, ok := profileRegistry[p]; ok && effect.maxCPUs > 0 {                                                                                                                                   
              if cap == 0 || effect.maxCPUs < cap {                                                                                                                                                         
                  cap = effect.maxCPUs
              }                                                                                                                                                                                             
          }                                                 
      }
      return cap
  }

  3. No validation that override packages match known operators                                                                                                                                             
   
  ValidateOperatorOverrides only rejects empty strings. A typo in the package name (e.g. rhoods-operator instead of rhods-operator) silently creates an unused CatalogSource with no warning to the user.   
                                                            
  4. No test coverage                                                                                                                                                                                       
                                                            
  Validate, ValidateOperatorOverrides, ensureCatalogSources, and the override logic in installOperator have no unit tests. Table-driven tests for the validation functions would at minimum catch           
  regressions.                                              
                                                                                                                                                                                                            
  ---                                                       
  Minor
       
  - The hash[:8] truncation (32 bits) is fine for human-facing names; practical collision risk is negligible.
  - sourceType: grpc is correct for FBC index images.                                                                                                                                                       
  - The --catalog-source nfd=quay.io/... example in the docs is clear.                                                                                                                                      
                                                                                                                                                                                                            
  ---                                                                                                                                                                                                       
  Summary                                                   
         
  The approach is solid and the FBC mechanism (custom CatalogSource + channel override) is correct. The critical gap is #1 — without waiting for the CatalogSource to be READY, the RHOAI subscription will
  race against OLM catalog indexing and fail intermittently. The fix is straightforward using the waitForCRCondition pattern already in the codebase.                                                       
  ```

amastbau:
Hey Jose, just wanted to share that I tested this PR as part of [AIPCC-19538](https://redhat.atlassian.net/browse/AIPCC-19538) — validating mapt's `--profile ai` with RHOAI stable-3.x channel.

Used `--operator-channel rhods-operator=stable-3.x` and it worked perfectly — RHOAI 3.4.2 deployed on SNO in ~22 minutes with all prerequisites (ServiceMesh, Serverless, KServe) ready.

FWIW, this flag is needed for anyone wanting RHOAI 3.x since the default `stable` channel still gives 2.25.x.

Assisted-by: Claude Code

adrianriobo:
wow that is nice @amastbau , @jangel97 did you have the chance to check my comment? if you want we can create a follow up issue, go and merge this one WDYT?

jangel97:
> wow that is nice @amastbau , @jangel97 did you have the chance to check my comment? if you want we can create a follow up issue, go and merge this one WDYT?

@adrianriobo sounds good to me, we can merge it and I do follow-up MR next sprint to tackle the feedback

adrianriobo (approved):
LGTM

PR #810, opened 2026-05-20
fix(snc-profiles): Fix non reachable api while destroying SNC

### LINKED ISSUE #809: [BUG] Openshift snc resources destroy error when api is not reachable
We define a strategy to nto acitvely remove resources deployed on top of openshift snc instead we rely the destroy within the infra (ASG or instances) but sill during destroy it tries to check the state of resources with the cluster and if it is not reachable it fails.

adrianriobo:
Previously we set the delete of any resource deployed on top of SNC as deleted with the underlaying infra, but even in that case during destroy pulumi was trying to reach the SNC api to check the state for resources, and when it is unrechable it was failing the whole destroy operation, we are now forcing the DeleteUnrechable to true, so it bypass it even if it can not reach the cluster to check the state. 

Fixes #809


jangel97 (approved):

jangel97:
LGTM

PR #811, opened 2026-05-21
feat(ibmcloud): add IBM COS S3-compatible backend support

deekay2310:
Introduce IBM_COS_* env vars (access key, secret, endpoint, region) and remap them to AWS SDK env vars in Init() so the Pulumi S3 backend can authenticate to IBM Cloud Object Storage without conflicting with AWS credentials. Add DestroyStack with lock cleanup, CleanupState for post- destroy state removal, and --keep-state flag to ibm-power/ibm-z destroy commands.

### DIFF
--- a/pkg/provider/ibmcloud/constants/constants.go
+++ b/pkg/provider/ibmcloud/constants/constants.go
@@ -1,0 +1,8 @@
+package constants
+
+const (
+	EnvIBMCosAccessKeyID     = "IBM_COS_ACCESS_KEY_ID"
+	EnvIBMCosSecretAccessKey = "IBM_COS_SECRET_ACCESS_KEY"
+	EnvIBMCosEndpoint        = "IBM_COS_ENDPOINT"
+	EnvIBMCosRegion          = "IBM_COS_REGION"
+)
--- a/pkg/provider/ibmcloud/constants/constants.go
+++ b/pkg/provider/ibmcloud/constants/constants.go
@@ -1,0 +1,9 @@
+package constants
+
+const (
+	EnvIBMCloudAccount       = "IBMCLOUD_ACCOUNT"
+	EnvIBMCloudAPIKey        = "IBMCLOUD_API_KEY"
+	EnvIBMCosAccessKeyID     = "IBMCLOUD_COS_ACCESS_KEY_ID"
+	EnvIBMCosSecretAccessKey = "IBMCLOUD_COS_SECRET_ACCESS_KEY"
+	EnvIBMCosEndpoint        = "IBMCLOUD_COS_ENDPOINT"
+)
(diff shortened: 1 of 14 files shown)


pkg/provider/ibmcloud/ibmcloud.go
```diff
@@ -50,6 +59,119 @@ var (
...
+func parseS3BackedURL(mCtx *mc.Context) (*string, *string, error) {
+	if !strings.HasPrefix(mCtx.BackedURL(), "s3://") {
+		return nil, nil, fmt.Errorf("invalid S3 URI: must start with s3://")
+	}
+	u, err := url.Parse(mCtx.BackedURL())
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to parse S3 URI: %w", err)
+	}
+	key := strings.TrimPrefix(u.Path, "/")
+	return &u.Host, &key, nil
```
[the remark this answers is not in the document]
adrianriobo: @deekay2310 can you tackle this one?

pkg/provider/ibmcloud/constants/constants.go
```diff
@@ -0,0 +1,8 @@
+package constants
+
+const (
+	EnvIBMCosAccessKeyID     = "IBM_COS_ACCESS_KEY_ID"
```
adrianriobo: We already got IBMCLOUD_ACCOUNT and IBMCLOUD_API_KEY and IC_REGION which are used by the natively by the provider, should we keep those I mean use them instead ? Also what is the value for the endpoint?

pkg/provider/ibmcloud/ibmcloud.go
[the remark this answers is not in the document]
deekay2310: Fixed it. Applied the same fix to the AWS provider's copy of the function as well.

pkg/provider/ibmcloud/constants/constants.go
deekay2310: Since these are HMAC keys (access key + secret key) and exist specifically for S3-compatible API access, they are separate from IAM API keys and I think we should keep them separate. For region, it should be okay to use IC_REGION.

Re: endpoint - value needs to be set manually right now. Should I look into auto-constructing it?

deekay2310: Made some changes [here](https://github.com/redhat-developer/mapt/pull/811/changes/6471f3e356968e95d8a469b1090fdd2ca5e72dcb).

adrianriobo:
@deekay2310 can you add sample of usage within the docs for IBM also the squash commits and I think we are good to go

adrianriobo:
NVM doc is all good, once you tested with last artifacts squash all commits then I will LGTM

deekay2310:
The COS Backend connection worked and Pulumi connected to COS successfully. Just going to need to harden how we delete the resources.

adrianriobo (approved):
LGTM

PR #823, opened 2026-06-04
fix(azure): set DiskControllerType in StorageProfile for RHEL AI VMs

rishupk:
RHEL AI VMs on certain instance families (`Standard_L8aos_v4` and others) fail at boot because Azure infers NVMe as the disk controller type, but the gallery images only work with SCSI. This sets `DiskControllerType: "SCSI"` in `StorageProfile` through a new field on `ImageReference`. It also reads the disk controller types from the gallery image definition and filters compute sizes at allocation — spot and non-spot — so an incompatible size fails with a clear error before Azure ever sees the request.

Co-Authored-By: Claude <Sonnet 4.6> <noreply@anthropic.com>


rishupk:

PR #824, opened 2026-06-04
fix(azure): harden NVMe filter and credential handling for compute-sizes

rishupk:
Azure L-series VMs report `MaxResourceVolumeMB = 0` but track NVMe storage under a separate capability, `NvmeDiskSizeInMiB`, which the local storage filter never read. `Standard_L8aos_v4` was passing the filter and landing in RHEL AI spot selection. Added `NvmeDiskSizeInMiB` parsing and `FilterNoLocalStorageSizes()` for the explicit `--compute-sizes` path.

E2E testing with `--compute-sizes` exposed two additional issues fixed here:

- **NVMe filter too aggressive**: `FilterNoLocalStorageSizes` rejected any VM with `MaxResourceVolumeMB > 0` (temp disk), including `Standard_NC64as_T4_v3` which works fine for RHEL AI. Relaxed to only reject NVMe-only local storage (L-series). Temp disks are ephemeral scratch space and don't interfere with RHEL AI's OS disk.

- **Credential env var mapping**: The Azure SDK reads `AZURE_TENANT_ID` / `AZURE_SUBSCRIPTION_ID` / etc from the environment, but callers typically set `ARM_*` vars. The `ARM_*` → `AZURE_*` mapping only happened in `provider.Init()`, which runs after `FilterNoLocalStorageSizes` — causing credential failures on the `--compute-sizes` path. Added `ensureAzureEnvs()` before all `NewDefaultAzureCredential` calls, and centralized subscription ID reads through a `SubscriptionID()` helper.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/pkg/provider/azure/data/compute-request.go
+++ b/pkg/provider/azure/data/compute-request.go
@@ -2,7 +2,6 @@ package data
 
 import (
 	"context"
-	"os"
 	"regexp"
 	"slices"
 	"strconv"
@@ -12,6 +11,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7"
 	cr "github.com/redhat-developer/mapt/pkg/provider/api/compute-request"
+	"github.com/redhat-developer/mapt/pkg/util/logging"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,10 @@
+{
```
adrianriobo: Recall remove this file from PR, that is weird I though we have in .gitignore

adrianriobo (changes requested):

.claude/settings.local.json
rishupk: oh my mistake, removing it

pkg/provider/azure/data/compute-request.go
```diff
@@ -144,17 +212,17 @@ func (vm *virtualMachine) hypervGen2Supported() bool {
...
-func (vm *virtualMachine) emptyDiskSupported() bool {
-	return vm.MaxResourceVolumeMB == 0
+func (vm *virtualMachine) noLocalStorageAttached() bool {
+	return vm.MaxResourceVolumeMB == 0 && vm.NvmeDiskSizeInMiB == 0
 }
 
 func (vm *virtualMachine) baseFeaturesSupported() bool {
 	return vm.AcceleratedNetworkingEnabled && vm.PremiumIO && vm.EncryptionAtHostSupported &&
-		vm.emptyDiskSupported() && vm.hypervGen2Supported()
+		vm.noLocalStorageAttached() && vm.hypervGen2Supported()
```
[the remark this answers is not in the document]
rishupk: `emptyDiskSupported()` has been inside `baseFeaturesSupported()` since the file was first created (commit c6b2de19d) — filtering out local-storage VMs from auto-selection was always a global constraint, not RHEL AI-specific. The only change here is closing the NVMe loophole: L-series VMs report `MaxResourceVolumeMB = 0` but expose storage under `NvmeDiskSizeInMiB`, so the old check missed them. No Azure action in the repo references L-series VMs, and mapt's auto-selection is designed around managed disks, so broadening the filter to catch NVMe local storage is consistent with the original intent.

rishupk:

PR #825, opened 2026-06-04
fix(azure/rhel-ai): add GPU instance guardrails

rishupk:
RHEL AI on Azure could end up on a non-GPU instance with no error. VM boots, vllm never starts, no obvious failure — just a machine that won't run the workload.

Two things changed. The Azure VM SKU filter now reads the `GPUs` capability from Resource SKU, and the RHEL AI action defaults `ComputeRequest.GPUs` to 1 before the allocation step runs — keeps ND/NC-series in scope for spot and auto-select. If someone passes `--compute-sizes` with non-GPU types, we log a warning.


rishupk:

PR #830, opened 2026-06-09
fix(ibmcloud): preserve path component in COS backend URL

### LINKED ISSUE #829: [Bug] Ibm Cloud backed url with COS only uses bucket but not the whole path
When running IBM CLoud using COS as backed storage and using a path only the bucket is picked to store the state:

```
...
--backed-url s3://bucket/path
...
```

We want to store the state under the path folder in the bucket

### LINKED ISSUE #832: [Issue] GitLab runners randomly fail while resolving DNS for https://github.com/pytorch/pytorch
GitLab runners for s390x randomly fail while resolving DNS for https://github.com/pytorch/pytorch

--- Cloning PyTorch repository ---
Cloning into '/pytorch'...
fatal: unable to access 'https://github.com/pytorch/pytorch/': Could not resolve host: github.com

The issue is flaky and resolves after a few re-runs. The issue happens 3/5 times.

adrianriobo:
When a backed URL includes a path (e.g. s3://bucket/some/path), the previous code discarded the path and stored Pulumi state at the bucket root. Refactor extractBucket into extractBucketAndPath so the path is carried through to PULUMI_BACKEND_URL.

Fixes #829 
Fixes #832

### DIFF
--- a/pkg/manager/context/context.go
+++ b/pkg/manager/context/context.go
@@ -191,7 +191,7 @@ func manageIntegration(c *Context, ca *ContextArgs) error {
 		cirrus.Init(ca.CirrusPWArgs)
 	}
 	if ca.GLRunnerArgs != nil {
-		ca.GLRunnerArgs.Name = c.RunID()
+		ca.GLRunnerArgs.Name = c.ProjectName()
 		gitlab.Init(ca.GLRunnerArgs)
 	}
 	return nil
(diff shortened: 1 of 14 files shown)


pkg/manager/context/context.go
deekay2310, line 186: Should this also be changed to c.ProjectName() just like GLRunnerArgs.name?

deekay2310 (commented):
Apart from my one comment regarding `context.go`, we should also address CodeRabbit's comments and I think this PR will be ready to merge.

pkg/manager/context/context.go
adrianriobo: yeah I just found out  using the RunID as name is not really helpful so make it with the project name clarifies it, but as this change is GH related can you change it on your PR?

deekay2310: Sounds good! I'll look into it

ppitonak (approved):

PR #831, opened 2026-06-09
feat(ibmcloud): add GitHub Actions runner support for IBM Power and IBM Z

deekay2310:
## Summary

Adds GitHub Actions self-hosted runner support for IBM Power (ppc64le) and IBM Z (s390x). Extends `mapt ibmcloud ibm-power create` and `mapt ibmcloud ibm-z create` with `--ghactions-runner` to provision VMs that auto-register as ephemeral GitHub Actions runners.

Since no official runner binaries exist for ppc64le/s390x, arch-specific cloud-init snippets clone [action-runner-image-pz](https://github.com/aipcc-cicd/action-runner-image-pz) and build the runner from source on the target VM.

### Key changes

- **Runner provisioning**: new cloud-init snippets for ppc64le and s390x that build the runner, configure it with `--ephemeral`, and start it via `nohup ./run.sh`
- **Auto token generation**: `--ghactions-runner-token` is auto-generated via GitHub API if not provided
- **System type default** (Power): defaults to `e1080` system type
- **Volume mount timeout** (Power): 15-minute timeout on data volume attachment with fallback to root filesystem
- **Fatal destroyForRetry**: partial stack cleanup failure is now fatal to prevent orphaned resource accumulation
- **IBM COS S3 backend**: Pulumi state stored in IBM Cloud Object Storage via HMAC-authenticated S3-compatible API
- **CI**: smoke test workflow for s390x runner

### E2E test results

| Arch | Profile/SysType | Zone | Result |
|---|---|---|---|
| ppc64le | e1080 | dal14 | PASS — [run 28071744390](https://github.com/deekay2310/mapt/actions/runs/28071744390) |
| s390x | mz2-16x128 | us-south-2 | PASS — [run 28072488599](https://github.com/deekay2310/mapt/actions/runs/28072488599) |

Both architectures: provision → cloud-init → runner build → GitHub registration → smoke test job → clean destroy verified.

### DIFF
--- a/cmd/mapt/cmd/aws/hosts/fedora.go
+++ b/cmd/mapt/cmd/aws/hosts/fedora.go
@@ -56,7 +56,7 @@ func getFedoraCreate() *cobra.Command {
 					Debug:         viper.IsSet(params.Debug),
 					DebugLevel:    viper.GetUint(params.DebugLevel),
 					CirrusPWArgs:  params.CirrusPersistentWorkerArgs(),
-					GHRunnerArgs:  params.GithubRunnerArgs(),
+					GHRunnerArgs:  params.GithubRunnerArgs(params.LinuxGithubArch()),
 					GLRunnerArgs:  params.GitLabRunnerArgs(params.LinuxGitLabArch()),
 					Tags:          viper.GetStringMapString(params.Tags),
 				},
--- a/cmd/mapt/cmd/ibmcloud/hosts/ibm-power.go
+++ b/cmd/mapt/cmd/ibmcloud/hosts/ibm-power.go
@@ -2,6 +2,7 @@ package hosts
 
 import (
 	"github.com/redhat-developer/mapt/cmd/mapt/cmd/params"
+	"github.com/redhat-developer/mapt/pkg/integrations/github"
 	"github.com/redhat-developer/mapt/pkg/integrations/gitlab"
 	maptContext "github.com/redhat-developer/mapt/pkg/manager/context"
 	ibmpower "github.com/redhat-developer/mapt/pkg/provider/ibmcloud/action/ibm-power"
@@ -52,7 +53,7 @@ func ibmPowerCreate() *cobra.Command {
 					Debug:         viper.IsSet(params.Debug),
 					DebugLevel:    viper.GetUint(params.DebugLevel),
 					CirrusPWArgs:  params.CirrusPersistentWorkerArgs(),
-					GHRunnerArgs:  params.GithubRunnerArgs(),
+					GHRunnerArgs:  params.GithubRunnerArgs(&github.Ppc64le),
 					GLRunnerArgs:  params.GitLabRunnerArgs(&gitlab.Ppc64le),
 					Tags:          viper.GetStringMapString(params.Tags),
 				},
--- a/cmd/mapt/cmd/ibmcloud/hosts/ibm-z.go
+++ b/cmd/mapt/cmd/ibmcloud/hosts/ibm-z.go
@@ -2,6 +2,7 @@ package hosts
 
 import (
 	"github.com/redhat-developer/mapt/cmd/mapt/cmd/params"
+	"github.com/redhat-developer/mapt/pkg/integrations/github"
 	"github.com/redhat-developer/mapt/pkg/integrations/gitlab"
 	maptContext "github.com/redhat-developer/mapt/pkg/manager/context"
 	ibmz "github.com/redhat-developer/mapt/pkg/provider/ibmcloud/action/ibm-z"
@@ -52,7 +53,7 @@ func ibmZCreate() *cobra.Command {
 					Debug:         viper.IsSet(params.Debug),
 					DebugLevel:    viper.GetUint(params.DebugLevel),
 					CirrusPWArgs:  params.CirrusPersistentWorkerArgs(),
-					GHRunnerArgs:  params.GithubRunnerArgs(),
+					GHRunnerArgs:  params.GithubRunnerArgs(&github.S390x),
 					GLRunnerArgs:  params.GitLabRunnerArgs(&gitlab.S390x),
 					Tags:          viper.GetStringMapString(params.Tags),
 				},
--- a/cmd/mapt/cmd/params/params.go
+++ b/cmd/mapt/cmd/params/params.go
@@ -1,6 +1,8 @@
 package params
 
 import (
+	"os"
+
 	"github.com/redhat-developer/mapt/pkg/integrations/cirrus"
 	"github.com/redhat-developer/mapt/pkg/integrations/github"
 	"github.com/redhat-developer/mapt/pkg/integrations/gitlab"
@@ -155,8 +157,8 @@ const (
 	PIProcTypeDesc       string  = "PowerVS processor type (shared, dedicated, capped)"
 	PIProcTypeDefault    string  = "shared"
 	PISysType            string  = "pi-sys-type"
-	PISysTypeDesc        string  = "PowerVS system type (s922, s1022, e880, e980)"
-	PISysTypeDefault     string  = "s1022"
+	PISysTypeDesc        string  = "PowerVS system type (e.g. e1080, s1022, s1122)"
+	PISysTypeDefault     string  = "e1080"
 	PIStorageType        string  = "pi-storage-type"
 	PIStorageTypeDesc    string  = "PowerVS storage tier for instance and data volume (tier1, tier3)"
 	PIStorageTypeDefault string  = "tier1"
@@ -282,18 +284,38 @@ func AddGHActionsFlags(fs *pflag.FlagSet) {
 	fs.StringSlice(ghActionsRunnerLabels, nil, GHActionsRunnerLabelsDesc)
 }
 
-func GithubRunnerArgs() *github.GithubRunnerArgs {
-	if viper.IsSet(ghActionsRunnerToken) {
-		return &github.GithubRunnerArgs{
-			Token:    viper.GetString(ghActionsRunnerToken),
-			RepoURL:  viper.GetString(ghActionsRunnerRepo),
-			Labels:   viper.GetStringSlice(ghActionsRunnerLabels),
-			Platform: &github.Linux,
-			Arch: linuxArchAsGithubActionsArch(
-				viper.GetString(LinuxArch)),
-		}
+func GithubRunnerArgs(arch *github.Arch) *github.GithubRunnerArgs {
+	token := viper.GetString(ghActionsRunnerToken)
+	repoURL := viper.GetString(ghActionsRunnerRepo)
+	pat := os.Getenv("GITHUB_TOKEN")
+
+	if token == "" && pat == "" {
+		return nil
 	}
-	return nil
+
+	if repoURL == "" {
+		logging.Error("--ghactions-runner-repo is required for GitHub Actions runner setup")
+		return nil
+	}
+
+	if token == "" {
+		logging.Info("no --ghactions-runner-token provided, auto-generating from GITHUB_TOKEN")
+		var err error
+		token, err = github.GenerateRegistrationToken(pat, repoURL)
+		if err != nil {
+			logging.Errorf("failed to auto-generate runner registration token: %v", err)
+			return nil
+		}
+		logging.Info("runner registration token generated successfully")
+	}
+
+	return &github.GithubRunnerArgs{
+		Token:    token,
+		RepoURL:  repoURL,
+		Labels:   viper.GetStringSlice(ghActionsRunnerLabels),
+		Platform: &github.Linux,
+		Arch:     arch,
+	}
 }
 
 func AddCirrusFlags(fs *pflag.FlagSet) {
@@ -357,10 +379,18 @@ func linuxArchAsCirrusArch(arch string) *cirrus.Arch {
 	return &cirrus.Arm64
 }
 
+func LinuxGithubArch() *github.Arch {
+	return linuxArchAsGithubActionsArch(viper.GetString(LinuxArch))
+}
+
 func linuxArchAsGithubActionsArch(arch string) *github.Arch {
 	switch arch {
 	case "x86_64":
 		return &github.Amd64
+	case "ppc64le":
+		return &github.Ppc64le
+	case "s390x":
+		return &github.S390x
 	}
 	return &github.Arm64
 }
--- a/pkg/integrations/github/api.go
+++ b/pkg/integrations/github/api.go
@@ -1,0 +1,68 @@
+package github
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+)
+
+type registrationTokenResponse struct {
+	Token     string `json:"token"`
+	ExpiresAt string `json:"expires_at"`
+}
+
+// GenerateRegistrationToken calls the GitHub API to create a short-lived
+// runner registration token for the given repository.
+// pat is a Personal Access Token with repo admin scope.
+// repoURL is in the form "owner/repo" or "https://github.com/owner/repo".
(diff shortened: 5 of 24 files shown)


cmd/mapt/cmd/params/params.go
```diff
@@ -278,18 +287,54 @@ func AddGHActionsFlags(fs *pflag.FlagSet) {
...
+		RunnerImageRepo: imageRepo,
+	}
+}
+
+func validateRunnerImageRepo(repo string) error {
+	if !strings.HasPrefix(repo, "https://") {
+		return fmt.Errorf("only HTTPS URLs are allowed, got: %s", repo)
 	}
 	return nil
 }
```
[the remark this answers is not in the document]
deekay2310: Hardcoding an allowlist of specific GitHub repos makes the tool inflexible for downstream users who fork or maintain their own runner image repos. An allowlist would need updating every time someone forks the repo which does not seem feasible to me.

pkg/provider/ibmcloud/data/pisystempools.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	return false
+}
+
+func poolAvailableMemory(pool *models.SystemPool) int64 {
+	if pool.MaxAvailable != nil && pool.MaxAvailable.Memory != nil {
+		return *pool.MaxAvailable.Memory
+	}
+	return 0
+}
```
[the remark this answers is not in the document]
deekay2310: Fixed — `poolAvailableMemory` now checks all three capacity sources (`MaxCoresAvailable`, `MaxMemoryAvailable`, `MaxAvailable`) consistent with `poolHasCapacity`. The change has been pushed.

pkg/provider/ibmcloud/action/ibm-power/cloud-config
```diff
@@ -11,25 +11,33 @@ write_files:
...
-        cp -a /home/. /mnt/home-tmp/
-        mount "$DATA_DEV" /home
-        cp -a /mnt/home-tmp/. /home/
-        rm -rf /mnt/home-tmp
-        echo "UUID=$UUID /home xfs defaults 0 2" >> /etc/fstab
-        mkdir -p /home/containers/storage
+        if [ -z "$DATA_DEV" ]; then
+          echo "WARNING: data volume not found after $((MAX_ATTEMPTS * 10))s, using root filesystem" >&2
+          mkdir -p /home/containers/storage
+        else
```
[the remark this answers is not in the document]
deekay2310: Not adding a sentinel file — the existing code already handles this:

1. **Successful mount path**: the volume is added to `/etc/fstab` (line 38), so on reboot `/home` is already a mountpoint and the `mountpoint -q /home` check (line 10) short-circuits before the loop.
2. **Fallback path (no volume)**: these are ephemeral runners (`--ephemeral`) that get destroyed after one job — a reboot where the loop re-runs is not a realistic scenario.

The 15-minute timeout we added is sufficient protection for the first-boot case.

pkg/manager/context/context.go
```diff
@@ -183,7 +183,7 @@ func addCommonTags(c *Context) {
 
 func manageIntegration(c *Context, ca *ContextArgs) error {
 	if ca.GHRunnerArgs != nil {
-		ca.GHRunnerArgs.Name = c.RunID()
+		ca.GHRunnerArgs.Name = c.ProjectName()
```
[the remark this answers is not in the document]
deekay2310: No change needed — `--ephemeral` and `--replace` serve different purposes here by design:

- **x86/macOS** (`snippet-linux.sh`, `snippet-darwin.sh`): persistent runners that survive across jobs, so `--replace` handles re-registration with the same name.
- **ppc64le/s390x** (`snippet-linux-ppc64le.sh`, `snippet-linux-s390x.sh`): ephemeral runners that auto-deregister after one job. `--replace` is unnecessary and actually incompatible with `--ephemeral` (GitHub docs: "the `--replace` flag is ignored when used with `--ephemeral`").

Naming conflicts are also not possible because the runner name comes from `c.ProjectName()` which includes a random hash suffix (e.g. `maptb5bdcacc`) — each provisioning run gets a unique name.

cmd/mapt/cmd/ibmcloud/hosts/ibm-power.go
```diff
@@ -43,6 +44,10 @@ func ibmPowerCreate() *cobra.Command {
 			if err := viper.BindPFlags(cmd.Flags()); err != nil {
 				return err
 			}
+			ghRunnerArgs := params.GithubRunnerArgs()
```
adrianriobo: You can do similar to GitlabRunners func 

params.GithubRunnerArgs() -> params.GithubRunnerArgs(&github.Ppc64le)

then you remove this lines and just

...
GHRunnerArgs:  params.GithubRunnerArgs(&github.Ppc64le),
...

cmd/mapt/cmd/ibmcloud/hosts/ibm-z.go
```diff
@@ -43,6 +44,10 @@ func ibmZCreate() *cobra.Command {
 			if err := viper.BindPFlags(cmd.Flags()); err != nil {
 				return err
 			}
+			ghRunnerArgs := params.GithubRunnerArgs()
```
adrianriobo: Same as previous comment

cmd/mapt/cmd/params/params.go
```diff
@@ -73,9 +77,14 @@ const (
 	CreateCmdName  string = "create"
 	DestroyCmdName string = "destroy"
 
-	ghActionsRunnerToken  string = "ghactions-runner-token"
-	ghActionsRunnerRepo   string = "ghactions-runner-repo"
-	ghActionsRunnerLabels string = "ghactions-runner-labels"
+	ghActionsRunnerToken    string = "ghactions-runner-token"
+	ghActionsRunnerRepo     string = "ghactions-runner-repo"
+	ghActionsRunnerLabels   string = "ghactions-runner-labels"
+	ghActionsRunnerImageRepo string = "ghactions-runner-image-repo"
```
adrianriobo: Probably we dont want to expose this...you can set it in the logic (not in cmd)  and also instead of download the zip pointing to main...use tags and download source from releases

cmd/mapt/cmd/params/params.go
```diff
@@ -290,18 +289,59 @@ func AddGHActionsFlags(fs *pflag.FlagSet) {
...
-		return &github.GithubRunnerArgs{
-			Token:    viper.GetString(ghActionsRunnerToken),
-			RepoURL:  viper.GetString(ghActionsRunnerRepo),
-			Labels:   viper.GetStringSlice(ghActionsRunnerLabels),
-			Platform: &github.Linux,
-			Arch: linuxArchAsGithubActionsArch(
-				viper.GetString(LinuxArch)),
+	token := viper.GetString(ghActionsRunnerToken)
+	repoURL := viper.GetString(ghActionsRunnerRepo)
+	pat := os.Getenv("GITHUB_TOKEN")
```
adrianriobo, line 290: what is this?

pkg/integrations/github/api.go
```diff
@@ -0,0 +1,68 @@
...
+	ownerRepo = strings.TrimPrefix(ownerRepo, "https://github.com/")
+	ownerRepo = strings.TrimPrefix(ownerRepo, "http://github.com/")
+	ownerRepo = strings.TrimSuffix(ownerRepo, "/")
+
+	parts := strings.Split(ownerRepo, "/")
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", fmt.Errorf("invalid repo format %q, expected owner/repo", repoURL)
+	}
+
+	url := fmt.Sprintf("https://api.github.com/repos/%s/%s/actions/runners/registration-token", parts[0], parts[1])
```
adrianriobo, line 32: Why do we use this? Can we use https://www.pulumi.com/registry/packages/github/api-docs/getactionsregistrationtoken/ instead, so this is tied to Pulumi cycle...actually probably this will fix the de registration on destroy ??

pkg/integrations/github/snippet-linux-ppc64le.sh
```diff
@@ -0,0 +1,115 @@
...
+        awk '!seen[$0]++' "$f" > "${f}.tmp" && mv "${f}.tmp" "$f"
+    fi
+done
+systemctl restart sshd 2>/dev/null || true
+
+echo "=== POST-REPAIR SSHD STATUS ===" >> /var/log/sshd-watchdog.log
+systemctl status sshd >> /var/log/sshd-watchdog.log 2>&1
+ss -tlnp | grep :22 >> /var/log/sshd-watchdog.log 2>&1
+
+# Upload diagnostics to COS so we can read them without SSH
```
adrianriobo: What is this for?

pkg/integrations/github/snippet-linux-ppc64le.sh
```diff
@@ -0,0 +1,115 @@
...
+    exit 1
+fi
+
+id -u runner &>/dev/null || useradd -m -s /bin/bash runner
+chown -R runner:runner /opt/runner-cache
+
+sudo -u runner bash -c '
+    cd /opt/runner-cache
+
+    ./config.sh \
```
adrianriobo, line 34: Can we explore options in config to only install the cli and set it up to join as runner?

pkg/provider/ibmcloud/action/ibm-power/ibm-power.go
```diff
@@ -113,36 +119,105 @@ func New(ctx *mc.ContextArgs, args *PWArgs) error {
...
+		Zone:            os.Getenv("IC_ZONE"),
+		ProcType:        args.ProcType,
+		PreferredType:   args.SysType,
+	})
+	if err != nil {
+		return fmt.Errorf("system type discovery failed: %w", err)
 	}
+
+	var lastErr error
+	for i, sysType := range sysTypes.Types {
```
adrianriobo: I think we may can rely on `e` series as default it should not hit capacity issues...if we see them may we can create a follow up to find a solution for all providers and archs as this problem is not affecting only this arch and provider. WDYT?

adrianriobo (changes requested):

cmd/mapt/cmd/params/params.go
deekay2310: Removed the `--ghactions-runner-image-repo` CLI flag and hardcoded the repo URL internally in `ghrunner.go`. For switching to tags/releases instead of cloning `main`, I can take care of that in a follow-up PR once the fork is moved into `aipcc-cicd` and we have tagged releases to point at. Does that work?

cmd/mapt/cmd/params/params.go
deekay2310: If the user doesn't provide `--ghactions-runner-token`, the code uses GITHUB_TOKEN from their environment to auto-generate a runner registration token via the GitHub API (GenerateRegistrationToken). This way users don't have to manually create a short-lived registration token before each run.

pkg/integrations/github/api.go
deekay2310: Using the Pulumi GitHub provider's `getActionsRegistrationToken` is a great idea, I was frankly not aware of it and went ahead with simplifying the short-lived token generation process through `GITHUB_TOKEN`. 

Since this looks like a significant change, can I create a follow-up PR for this as well?

pkg/integrations/github/snippet-linux-ppc64le.sh
deekay2310: Once the fork is moved into `aipcc-cicd` with tagged releases, we can publish pre-built binaries and simplify this to a download + configure like the `x86_64/arm64` snippet.

pkg/integrations/github/snippet-linux-ppc64le.sh
deekay2310: This was debug scaffolding (sshd watchdog, diagnostics logging, COS log upload) I used while investigating an issue where the upstream runner image build script breaks sshd by running chmod -R 777 /usr/share. Removed the diagnostics and kept only the sshd repair that's needed for the host to remain accessible after the build.

pkg/provider/ibmcloud/action/ibm-power/ibm-power.go
deekay2310: Sounds good to me. That'll also give us the chance to further improve it. Let me make those changes.

cmd/mapt/cmd/aws/hosts/fedora.go
```diff
@@ -56,7 +56,7 @@ func getFedoraCreate() *cobra.Command {
 					Debug:         viper.IsSet(params.Debug),
 					DebugLevel:    viper.GetUint(params.DebugLevel),
 					CirrusPWArgs:  params.CirrusPersistentWorkerArgs(),
-					GHRunnerArgs:  params.GithubRunnerArgs(),
+					GHRunnerArgs:  params.GithubRunnerArgs(nil),
```
adrianriobo: this should have the arch, in this case instead of fixed arch value it is coming from another flag `--arch`, this applies to almost all targets

adrianriobo (changes requested):

pkg/integrations/github/api.go
adrianriobo: sure just create a follow up (issue) from this comment and you can tackle it there

deekay2310: https://github.com/redhat-developer/mapt/issues/851#issue-4785731385

cmd/mapt/cmd/params/params.go
deekay2310: since the repo was moved into the `aipcc-cicd` org, I did a v2.0.0 release to be consumed in this section, which has now been implemented.

cmd/mapt/cmd/aws/hosts/fedora.go
deekay2310: fixed it

deekay2310:
## E2E Test Results — 2026-07-01 (latest image `ghcr.io/redhat-developer/mapt:pr-831`)

Provisioned both architectures from the freshly built container image (all CI checks passed), ran smoke-test workflows on the self-hosted runners.

### Provisioning

| Target | Result | Duration | Backend |
|--------|--------|----------|---------|
| **s390x** (ibm-z) | :white_check_mark: Success | ~2 min | S3 (IBM COS) |
| **ppc64le** (ibm-power) | :white_check_mark: Success | ~5 min | S3 (IBM COS) |

### Smoke Test Workflows

Both runners picked up and completed `workflow_dispatch` jobs successfully:

| Runner | Arch | OS | Disk | RAM | Docker | Podman | Workflow Run |
|--------|------|----|------|-----|--------|--------|--------------|
| `ibmz-gh` | `s390x` | Ubuntu 24.04.4 LTS | 246G (4%) | 125Gi | :x: (GPG key issue¹) | 4.9.3 | [28530981177](https://github.com/deekay2310/mapt/actions/runs/28530981177) |
| `ppc64-gh` | `ppc64le` | RHEL 9.6 | 99G (34%) | 92Gi | v28.0.4 | 5.4.0 | [28530984334](https://github.com/deekay2310/mapt/actions/runs/28530984334) |

### Known Issue

¹ Docker fails to install on s390x/Ubuntu 24.04 due to a Docker GPG key verification failure + `containerd.io` version `1.7.27-1` not available for s390x on noble. The runner image installer lacks a version-not-found fallback (unlike the RHEL installer). Tracked in [aipcc-cicd/action-runner-image-pz#2](https://github.com/aipcc-cicd/action-runner-image-pz/issues/2). Podman is available as a fallback.

### Summary

The full E2E flow is verified for both architectures: **provision → cloud-init → runner registration → workflow execution**. All latest refactoring commits (runner image pinned to `aipcc-cicd v2.0.0`, arch passed explicitly, capacity retry removed, runner image repo internalized) are working correctly.

adrianriobo (approved):
LGTM

PR #833, opened 2026-06-10
fix(rhel-ai): fix rhel-ai on demand EIP creation

adrianriobo:
previously we had craete load balancer fixed to true in rhel-ai (this is required for spot) but it should be false for on demand, this commit ensures it is created based on the type of provisioning


adrianriobo:
@ppitonak can you check / approve this is on line change and it finally fix the `rhel-ai` on demand

Tested and it works ok

ppitonak (approved):

PR #840, opened 2026-06-17
fix(azure): make location optional for spot allocation

### LINKED ISSUE #839: Mandatory parameter ARM_LOCATION_HOME for azure
I'm moving from version 0.9.6 to 0.14.1:

```
#!/bin/bash

VERSION=$1

podman run --rm \
    -v ${PWD}/workspace:/workspace:z \
    -e ARM_TENANT_ID='redacted' \
    -e ARM_SUBSCRIPTION_ID='redacted' \
    -e ARM_CLIENT_ID='redacte' \
    -e ARM_CLIENT_SECRET='redacted' \
    -e AZURE_STORAGE_ACCOUNT='redacted \
    -e AZURE_STORAGE_KEY='redacted' \
    -e BLOB='redacted' \
    quay.io/redhat-developer/mapt:v$VERSION azure \
       fedora create \
         --project-name fedora \
         --backed-url file:///workspace \
         --conn-details-output /workspace \
         --arch arm64 \
         --cpus 8 \
         --memory 64 \
         --spot \
         --spot-eviction-tolerance low \
         --tags pipelinerunName=mytest
```

0.9.6 work just fine
```
./create_azure_fedora 0.9.6
```

0.12.1 show a warning and some errors, but provision a machine just fine too
```
./create_azure_fedora 0.12.1
INFO missing default value for Azure Location, if needed it should set using ARM_LOCATION_NAME or AZURE_DEFAULTS_LOCATION 
...
ERRO GET https://management.azure.com/subscriptions/b0ad4737-8299-4c0a-9dd5-959cbcf8d81c/providers/Microsoft.Compute/locations/newzealandnorth/communityGalleries/Fedora-5e266ba4-2250-406d-adad-5d73860d958f/images/Fedora-Cloud-42-Arm64
--------------------------------------------------------------------------------
RESPONSE 404: 404 Not Found
ERROR CODE: NotFound
--------------------------------------------------------------------------------
{
  "error": {
    "code": "NotFound",
    "message": "please check if the gallery name 'Fedora-5e266ba4-2250-406d-adad-5d73860d958f' is valid and the gallery exists in 'NewZealandNorth' region."
  }
}

```

0.14.1 simply complains about those parameters:
```
./create_azure_fedora 0.14.1
DEBU running 'mapt azure fedora create'           
missing default value for Azure Location, set ARM_LOCATION_NAME or AZURE_DEFAULTS_LOCATION environment variable
```

Just adding anything
```
-e ARM_LOCATION_NAME=foo \
```

it provision correctly the machine, showing same NotFound errors as in version 0.12.1

rishupk:
Azure spot allocation finds its own location, so there's no reason to require one at init time. A fix in #826 (commit 7c5d50d) changed `DefaultHostingPlace()` to hard-error when no location env var is set — that worked fine for on-demand but also fired for spot, breaking workflows that never needed it (see #839).

This restores the original `nil, nil` return with an info log. The non-spot path in `allocation.go` already enforces location when it actually matters.

Fixes #839.

Co-Authored-By: Claude <Sonnet 4.6> <noreply@anthropic.com>

### DIFF
--- a/pkg/provider/azure/azure.go
+++ b/pkg/provider/azure/azure.go
@@ -22,10 +22,16 @@ var azIdentityEnvs = []string{
 	"AZURE_CLIENT_SECRET",
 }
 
-type Azure struct{}
+type Azure struct {
+	locationOptional bool
+}
 
 func Provider() *Azure {
 	return &Azure{}
+}
+
+func ProviderOptionalLocation() *Azure {
+	return &Azure{locationOptional: true}
 }
 
 func (a *Azure) Init(ctx context.Context, backedURL string) (string, error) {
@@ -34,6 +40,9 @@ func (a *Azure) Init(ctx context.Context, backedURL string) (string, error) {
 }
 
 func (a *Azure) DefaultHostingPlace() (*string, error) {
+	if a.locationOptional {
+		return nil, nil
+	}
 	hp := os.Getenv("ARM_LOCATION_NAME")
 	if len(hp) > 0 {
 		return &hp, nil
--- a/pkg/provider/azure/action/aks/aks.go
+++ b/pkg/provider/azure/action/aks/aks.go
@@ -59,7 +59,11 @@ func (r *aksRequest) validate() error {
 func Create(mCtxArgs *mc.ContextArgs, args *AKSArgs) (err error) {
 	logging.Debug("Creating AKS")
 	// Create mapt Context
-	mCtx, err := mc.Init(mCtxArgs, azure.Provider())
+	provider := azure.Provider()
+	if args.Spot != nil && args.Spot.Spot {
+		provider = azure.ProviderOptionalLocation()
+	}
+	mCtx, err := mc.Init(mCtxArgs, provider)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


pkg/provider/azure/azure.go
```diff
@@ -42,7 +42,8 @@ func (a *Azure) DefaultHostingPlace() (*string, error) {
 	if len(hp) > 0 {
 		return &hp, nil
 	}
-	return nil, fmt.Errorf("missing default value for Azure Location, set ARM_LOCATION_NAME or AZURE_DEFAULTS_LOCATION environment variable")
+	logging.Infof("missing default value for Azure Location, if needed it should set using ARM_LOCATION_NAME or AZURE_DEFAULTS_LOCATION")
+	return nil, nil
```
adrianriobo: I would expect a check in the options to see if the execution is for spot or not, 

* No spot the location is mandatory as it should be used as the target location for the on demand. So error
* Spot it will return nil

adrianriobo (changes requested):

pkg/provider/azure/azure.go
rishupk: Added `ProviderOptionalLocation()` — sets `locationOptional: true` on the `Azure` struct. `DefaultHostingPlace()` checks that flag before reading any env vars, so spot runs never get pinned to a location from the environment. Also applied this to all `Destroy` functions since teardown doesn't need a location and failing there would leave resources running.

adrianriobo (approved):
LGTM

Issue #847, opened 2026-06-26
RHELAI autstart (follow up)


adrianriobo:
After https://github.com/redhat-developer/mapt/pull/835 has been merged we want to follow up with some extensions:

- [ ] Extend logic to azure (+ azure tekton task)
- [ ] Introduce a https proxy (ideally we want to set a proxy) to offer vllm as https
- [ ] ExtendedPorts should apply to proxy port and should set the value we want (default may `6443`)
- [ ] Complete README.md
- [ ] Move setup to UserData
- [x] Add extended args,, ideally we want to pass something like [--enable-auto-tool-choice --tool-call-parser openai](https://docs.vllm.ai/en/stable/serving/integrations/claude_code/#starting-the-vllm-server)

PR #848, opened 2026-06-27
feat(aws): add --vllm-extra-args flag and auto GPU detection for RHEL AI

are-ces:
## Summary

Add `--vllm-extra-args` flag to `mapt aws rhel-ai create` for passing additional vLLM arguments, and auto-detect GPU count for tensor parallelism.

### `--vllm-extra-args`

A string flag that appends extra arguments to the RHAIIS vLLM `Exec=` line in install.conf. Allows configuring tool calling, context length, chat templates, and any future vLLM options without requiring new MAPT flags.

- If `--max-model-len` is included in the extra args, it replaces the default 4096 value
- If not included, the default 4096 is preserved and extra args are appended after it

### Auto GPU detection

Automatically detects the number of GPUs on the instance via `nvidia-smi` and sets `--tensor-parallel-size` accordingly. Works for any instance type without additional configuration.

**Usage:**

```bash
mapt aws rhel-ai create \
    --auto-start \
    --model meta-llama/Llama-3.1-8B-Instruct \
    --vllm-extra-args '--max-model-len 131072 --enable-auto-tool-choice --tool-call-parser llama3_json --chat-template /opt/app-root/template/tool_chat_template_llama3.1_json.jinja' \
    --expose-ports 8000 \
    ...
```

Updated Tekton task template with `vllm-extra-args` param.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/cmd/mapt/cmd/aws/hosts/rhelai.go
+++ b/cmd/mapt/cmd/aws/hosts/rhelai.go
@@ -72,6 +72,7 @@ func getRHELAICreate() *cobra.Command {
 					HFToken:          viper.GetString(params.RhelAIHFToken),
 					APIKey:           viper.GetString(params.RhelAIAPIKey),
 					AutoStart:        viper.IsSet(params.RhelAIAutoStart),
+					VLLMExtraArgs:    viper.GetString(params.RhelAIVLLMExtraArgs),
 					ExposePorts:      viper.GetIntSlice(params.RhelAIExposePorts),
 				})
 		},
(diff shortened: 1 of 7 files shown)


adrianriobo:
Hey @are-ces I created that follow up, issue for auto-start, https://github.com/redhat-developer/mapt/issues/847 I added one specific bullet for extended args, which basically you are trying here, but for other content...

I was wondering if it make sense (as I had in mind) just allow a flag at mapt level named (i.e.)  `--vllm-extra-args` and then basically you can pass them all together...

```
mapt aws rhel-ai \
     --auto-start \
     ....
     --vllm-extra-args '--tool-call-parser llama3_json --chat-template tool_chat_template_llama3.1_json.jinja  --max-model-len 16384'
```

The mapt flag may end up a bit long but that will allow us to move basically any vllm setup, as then we can use it as an openai compatible solution...

are-ces:
> Hey @are-ces I created that follow up, issue for auto-start, #847 I added one specific bullet for extended args, which basically you are trying here, but for other content...
> 
> I was wondering if it make sense (as I had in mind) just allow a flag at mapt level named (i.e.) `--vllm-extra-args` and then basically you can pass them all together...
> 
> ```
> mapt aws rhel-ai \
>      --auto-start \
>      ....
>      --vllm-extra-args '--tool-call-parser llama3_json --chat-template tool_chat_template_llama3.1_json.jinja  --max-model-len 16384'
> ```
> 
> The mapt flag may end up a bit long but that will allow us to move basically any vllm setup, as then we can use it as an openai compatible solution...

Yes I was thinking if this should be the right approach... let me update the PR

adrianriobo (approved):
LGTM

PR #856, opened 2026-07-08
fix(ibmcloud): remove PID limits to prevent EAGAIN during heavy builds

adrianriobo:
## Problem

On ppc64le (and s390x) runners, building wheel packages fails with:

\`\`\`
ccache: error: Failed to fork: Resource temporarily unavailable
gcc: error: posix_spawn: Operation not permitted
\`\`\`

Root cause: Pillow compiles 60+ C extension files. Each file spawns a
\`ccache → gcc → cc1 → as\` process chain (4 processes per file). Running
two concurrent jobs exhausts the container's cgroup \`pids.max\`, which
Podman inherits from the system default (~2048 on RHEL 9).

Three enforcement layers were all unset, each independently capable of
triggering the failure:

| Layer | Before | After |
|---|---|---|
| Podman \`pids_limit\` in \`containers.conf\` | system default | \`-1\` (unlimited) |
| Docker executor \`--docker-pids-limit\` | not set (inherits Podman default) | \`-1\` (unlimited) |
| Systemd \`TasksMax\` on \`gitlab-runner.service\` | systemd default (512–4096) | \`infinity\` |

## Changes

- **\`pkg/integrations/gitlab/snippet-linux.sh\`**: adds a \`pids_limit = -1\`
  block in \`/etc/containers/containers.conf\` (idempotent, covers all DNS
  code-path branches); adds \`--docker-pids-limit -1\` to \`gitlab-runner register\`
  so the Docker executor explicitly removes the pids cgroup on each job container.
- **\`pkg/provider/ibmcloud/action/ibm-power/cloud-config\`**: adds
  \`TasksMax=infinity\` to the \`gitlab-runner.service.d\` drop-in.
- **\`pkg/provider/ibmcloud/action/ibm-z/cloud-config\`**: same fix applied to
  IBM Z runners to prevent the same failure class on s390x.

## Test plan

- [x] Re-provision a ppc64le runner with these changes and trigger a Pillow wheel build
- [x] Confirm no EAGAIN / posix_spawn errors during C extension compilation
- [x] Verify s390x runner is unaffected (TasksMax=infinity is a no-op on non-constrained hosts)

### DIFF
--- a/pkg/integrations/gitlab/snippet-linux.sh
+++ b/pkg/integrations/gitlab/snippet-linux.sh
@@ -104,6 +104,21 @@ if [ ! -f /etc/containers/containers.conf ]; then
     sudo chmod 644 /etc/containers/containers.conf
 fi
 
+# Remove the per-container PID cap so compilation-heavy builds (e.g. Pillow
+# compiling 60+ C files through a ccache->gcc->cc1->as chain) do not exhaust
+# the cgroup pids.max and trigger EAGAIN on fork.
+if grep -q '^\[containers\]' /etc/containers/containers.conf; then
+    if awk '/^\[containers\]/{f=1;next} /^\[/{f=0} f && /^pids_limit/{found=1} END{exit !found}' \
+            /etc/containers/containers.conf; then
+        sudo sed -i 's|^pids_limit.*|pids_limit = -1|' /etc/containers/containers.conf
+    else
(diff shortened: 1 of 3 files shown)


Xaenalt:
So far, PID issue has not resurfaced, thanks!

adrianriobo:
I will merge this based in @Xaenalt comment

PR #864, opened 2026-07-22
fix(aws): forward AWS_SESSION_TOKEN to Pulumi provider

apodhrad:
This is needed when we use https://github.com/app-sre/rh-aws-saml-login


adrianriobo (approved):
LGTM

PR #865, opened 2026-07-23
fix(manager): skip Pulumi refresh during destroy to reduce memory usage

apodhrad:
The destroy path called stack.Refresh() before stack.Destroy(), which loaded Pulumi provider plugins twice and consumed ~26GB RAM for AWS stacks, causing OOM kills. Refresh is unnecessary for destroy since Pulumi handles stale state gracefully.

Extract credential setup from postStack() into setStackCredentials() and add getStackForDestroy() that sets credentials without refreshing.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -83,7 +83,7 @@ func DestroyStack(mCtx *mc.Context, targetStack Stack, opts ...ManagerOptions) e
 	logging.Debugf("destroying stack %s", targetStack.StackName)
 	ctx := mCtx.Context()
 
-	objectStack, err := getStack(ctx, mCtx, targetStack)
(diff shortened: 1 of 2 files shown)


apodhrad:
Provision and destroy needs to be properly tested before the merge. Thus, this is still a draft PR.

adrianriobo:
Tested and it works

adrianriobo (approved):
LGTM

apodhrad:
Hi @adrianriobo , I've removed the DRAFT prefix, so once you verify at other one more combination other that `aws openshift-snc` then you can merge it.

adrianriobo:
Done tested for Fedora and Windows Desktop (azure) all ok

PR #885, opened 2026-08-14
fix(azure): add Retail Prices API fallback for spot pricing

adrianriobo:
## Summary

- When the Azure Resource Graph `SpotResources` query returns no data, fall back to the public [Azure Retail Prices API](https://prices.azure.com/api/retail/prices) which requires no authentication
- Filters spot SKUs server-side by location and `contains(skuName, 'Spot')` (the API's OData subset does not support `or`, so compute-size filtering is done client-side)
- Handles pagination via `NextPageLink`
- Two-pass OS matching: first tries OS-specific spot SKUs; if none exist for the requested compute sizes (e.g. Windows spot not yet listed for newer v6/v7 VM series), falls back to any spot SKU scaled by `windowsOSPremiumFactor` (1.6×) to account for the Windows licensing fee absent from Linux spot prices

## Test plan

- [ ] Linux spot provisioning: Resource Graph path still works when data is available
- [ ] Linux spot provisioning: Retail API fallback returns correct pricing when Resource Graph returns empty
- [ ] Windows spot provisioning: Retail API returns Windows-specific SKU price when available
- [ ] Windows spot provisioning: OS-agnostic fallback with Windows premium factor produces a bid price that clears the actual Windows spot price (avoids 409 `OperationNotAllowed`)

### DIFF
--- a/pkg/provider/azure/data/spot.go
+++ b/pkg/provider/azure/data/spot.go
@@ -393,6 +393,9 @@ func checkSpotPricing(ctx context.Context, mCtx *mc.Context, locations []string,
 		}
 		results = append(results, rStruct)
 	}
+	if len(results) == 0 {
+		return checkSpotPricingRetailAPI(ctx, mCtx, locations, args)
+	}
 	return utilSlices.Split(
 		results,
 		func(s spotPricingResult) string {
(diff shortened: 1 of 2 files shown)


serbangeorge-m:
I managed to create an instance with this fix

praveenkumar:
@adrianriobo since it is tested by @serbangeorge-m and @anjannath  can we merge it and do a new release which will unblock CI runs?

adrianriobo:
Merged, though the issue should be fixed now, I checked from Azure console and graph query is returning data now again

PR #890, opened 2026-08-19
fix(tkn): make ocp-snc-credentials secret optional in infra-aws-ocp-snc

ppitonak:
## Summary

- \`infra-aws-ocp-snc.yaml\`'s \`pull-secret\` (from \`secret-ocp-snc-credentials\`) is only read during the \`create\` operation, but the volume mount was unconditional and the param had no default, so \`destroy\` calls fail to schedule unless a secret with that exact name already exists in the cluster.
- Marks the \`ocp-snc-credentials\` volume's secret as \`optional: true\` and defaults the \`secret-ocp-snc-credentials\` param to \`"''"\`, so \`destroy\` can run without requiring an unrelated secret to exist.

### DIFF
--- a/tkn/infra-aws-ocp-snc.yaml
+++ b/tkn/infra-aws-ocp-snc.yaml
@@ -22,6 +22,7 @@ spec:
     - name: ocp-snc-credentials
       secret:
         secretName: $(params.secret-ocp-snc-credentials)
+        optional: true
     - name: cluster-info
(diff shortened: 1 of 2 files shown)


adrianriobo:
@ppitonak missing the template file

ppitonak:
Both issues addressed

adrianriobo (approved):
LGTM

PR #891, opened 2026-08-19
fix(aws): tag IAM role/instance-profile and spot ASG with mapt resource tags

ppitonak:
The EC2 role, instance profile, spot Auto Scaling Group, and on-demand instances' root EBS volumes were untagged, making orphaned resources (from interrupted destroys) hard to identify and clean up. Also fixes a naming typo (instance-profie -> instance-profile).


adrianriobo (approved):
LGTM

PR #892, opened 2026-08-20
fix(otel): migrate to OTLP header-based authentication

adrianriobo:
## Summary

- Remove the interim resource-attribute token injection (`com.redhat.otel.auth_token`) from the otelcol-contrib collector config
- Add a standard `Authorization: Bearer` header on the `otlphttp` exporter for authenticated OTLP delivery
- Aligns with internal Enterprise Security Standard (ESS) requirements for authenticated OTLP data delivery

The configured endpoint default was already correct; no endpoint changes needed in code. Any secrets storing the old endpoint URL with an explicit port should be updated to use the new authenticated endpoint.

## Test plan

- [x] Deploy a VM/instance with `--otel-app-code` and `--otel-auth-token` set
- [x] Verify otelcol-contrib service starts and the `Authorization` header is present in outbound OTLP requests to the configured endpoint
- [ ] Confirm logs appear in the expected Splunk index

### DIFF
--- a/pkg/integrations/otelcol/snippet-linux.sh
+++ b/pkg/integrations/otelcol/snippet-linux.sh
@@ -182,9 +182,6 @@ processors:
       - key: appcode
         value: "{{.AppCode}}"
         action: upsert
-      - key: com.redhat.otel.auth_token
-        value: "${env:OTEL_AUTH_TOKEN}"
-        action: upsert
       - key: arch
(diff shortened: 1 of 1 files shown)


adrianriobo:
Tested with ppc64 and it is working as expected

anjannath (approved):
LGTM

PR #893, opened 2026-08-20
Make improvements

albfan:
Changes to make easier to jump into mapt development

- Set back install as default target
- Avoid to update your go version to match project


adrianriobo (approved):
LGTM

PR #898, opened 2026-09-07
fix(aws): wait for EIP association before readiness

ppitonak:
## Summary

- retain the on-demand EIP association resource
- make readiness and subsequent remote commands depend on the association
- prevent SSH retries from masking an `AssociateAddress` provisioning failure

## Context

When `ec2:AssociateAddress` fails, Pulumi currently starts the remote readiness command anyway because it depends only on the instance and EIP resources. The command targets the unattached EIP and reports `Dial N/inf failed: retrying` until its resource timeout.

## Testing

- `git diff --check`
- Go tests not run locally because the Go toolchain is unavailable in the workspace

### DIFF
--- a/pkg/provider/aws/modules/ec2/compute/compute.go
+++ b/pkg/provider/aws/modules/ec2/compute/compute.go
@@ -92,15 +92,16 @@ func (r *ComputeRequest) NewCompute(ctx *pulumi.Context) (*Compute, error) {
 			Eip:              r.Eip,
 			Dependencies:     []pulumi.Resource{asg, r.LB, r.Eip}}, err
 	}
-	i, err := r.onDemandInstance(ctx)
+	i, eipAssociation, err := r.onDemandInstance(ctx)
 	return &Compute{
 		Instance:     i,
 		Eip:          r.Eip,
-		Dependencies: []pulumi.Resource{i, r.Eip}}, err
+		// Commands use the EIP, so they must not start before it is attached.
+		Dependencies: []pulumi.Resource{eipAssociation}}, err
 }
 
 // Create on demand instance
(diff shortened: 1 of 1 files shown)


pkg/provider/aws/modules/ec2/compute/compute.go
```diff
@@ -92,15 +92,16 @@ func (r *ComputeRequest) NewCompute(ctx *pulumi.Context) (*Compute, error) {
...
 			Dependencies:     []pulumi.Resource{asg, r.LB, r.Eip}}, err
 	}
-	i, err := r.onDemandInstance(ctx)
+	i, eipAssociation, err := r.onDemandInstance(ctx)
 	return &Compute{
 		Instance:     i,
 		Eip:          r.Eip,
-		Dependencies: []pulumi.Resource{i, r.Eip}}, err
+		// Commands use the EIP, so they must not start before it is attached.
+		Dependencies: []pulumi.Resource{i, r.Eip, eipAssociation}}, err
```
adrianriobo: I am wondering after adding the association would it make sense to remove the `r.Eip`? that should be redundant?

jskladan:
I've ran a couple of tests with the `ghcr.io/redhat-developer/mapt:pr-898` image, and apart from one random-feeling hang on a `destroy` call, this seems to be working well.

pkg/provider/aws/modules/ec2/compute/compute.go
ppitonak: You are right, let me fix it

ppitonak: Done, now it waits for eipAssociation which guarantees both the instance and EIP to exist first

adrianriobo (approved):
LGTM

PR #899, opened 2026-09-07
fix(aws): skip S3 cleanup for local backends

ppitonak:
## Summary

- skip AWS state-object cleanup for non-S3 Pulumi backends
- avoid warning that a valid `file://` backend is an invalid S3 URI
- preserve parsing and warnings for malformed `s3://` backend URLs

## Testing

- `git diff --check`
- verified commit GPG signature
- Go tests not run because the Go toolchain is unavailable in the workspace


adrianriobo (approved):
LGTM

PR #900, opened 2026-09-08
feat(ibmcloud): add Kind cluster provisioning action

adrianriobo:
## Summary

- Provisions a single-node Kind (Kubernetes in Docker) cluster on an Ubuntu 24.04 x86_64 VM on IBM Cloud VPC, with a floating IP so the Kind API server certificate includes it as a SAN for direct `kubectl` access
- New `pkg/provider/ibmcloud/action/kind` package: VPC networking, SSH key, security group rules (6443/8888/9443 + configurable extra port mappings), MIME-wrapped cloud-init user data, and kubeconfig retrieval via remote commands
- IBM Cloud VPC compute profile selector (`pkg/provider/ibmcloud/data`): queries `ListInstanceProfiles` and picks the smallest profile satisfying requested CPUs/memory/arch/family constraints
- CLI commands: `mapt ibmcloud kind create/destroy` with `--cpus`, `--memory`, `--disk-size`, `--version`, `--extra-port-mappings` flags
- Tekton task + template (`infra-ibmcloud-kind`): IBM Cloud credentials secret, `zone` param exported as `IC_ZONE`, result secret contains `host`, `username`, `id_rsa`, and `kubeconfig`
- Documentation: `docs/ibmcloud/kind.md`

## Test plan

- [ ] `mapt ibmcloud kind create` provisions VM, deploys Kind cluster, and exports kubeconfig pointing to the floating IP
- [x] `kubectl get nodes` via exported kubeconfig connects successfully
- [x] `mapt ibmcloud kind destroy` tears down all resources cleanly
- [x] Compute profile selector resolves correct profile when `--cpus`/`--memory` flags are used
- [ ] Extra port mappings open correct security group rules and map into Kind config
- [ ] Tekton task `infra-ibmcloud-kind` creates/destroys cluster and populates result secret with kubeconfig


tkn/template/infra-ibmcloud-kind.yaml
```diff
@@ -0,0 +1,318 @@
...
+          else
+            echo "${cred:0:1}***${cred: -1}"
+          fi
+        }
+
+        # Credentials - set these BEFORE enabling debug mode
+        export IBMCLOUD_API_KEY=$(cat /opt/ibmcloud-credentials/api-key)
+        export IC_REGION=$(cat /opt/ibmcloud-credentials/region)
+        export IBMCLOUD_COS_ACCESS_KEY_ID=$(cat /opt/ibmcloud-credentials/cos-access-key)
+        export IBMCLOUD_COS_SECRET_ACCESS_KEY=$(cat /opt/ibmcloud-credentials/cos-secret-key)
```
deekay2310, line 191: Nit: `docs/ibmcloud/kind.md` lists `IBMCLOUD_ACCOUNT` as required, but it's not in the secret schema and not exported here. Should the doc be relaxed, or does the task need it?

deekay2310:
Reviewed the non-vendor source files (Go code, CLI, cloud-config, Tekton task/template, docs, Makefile). Vendor changes are the pulumi-ibmcloud SDK bump to v0.0.13 and transitive dependency updates, skipped those.

Everything looks good. Code follows existing codebase patterns, compute profile selector is well-structured, and the Tekton task is consistent with the Gaudi task. No blockers.

Left one minor inline comment about `IBMCLOUD_ACCOUNT` in the Tekton template.

adrianriobo:
@deekay2310 thx for quick review, actually I asked to use your tekon task as reference, when I run this test locally I do need the IBMCLOUD_ACCOUNT.... so not sure if the issue is we are missing it in both Taks (so in secrets format expectation as well)

Did you test the task?

tkn/template/infra-ibmcloud-kind.yaml
deekay2310: Ran `mapt ibmcloud ibm-gaudi create` locally with real creds but `IBMCLOUD_ACCOUNT` unset (verified it's not in the env). It got through creating ResourceGroup, VPC, Subnet, SecurityGroup, SSH Key, FloatingIP, and PublicGateway with no issues. Killed it before the instance finished spinning up. The API key alone is enough since it's already scoped to an account.

Grepping the code, `IBMCLOUD_ACCOUNT` is only consumed in the PowerVS paths (`piimages.go` and `power.go`), not VPC.

Also the run printed a deprecation warning on the SSH security group rule: `property "tcp" is deprecated, use 'protocol', 'code', and 'type' instead`, which is exactly what the `network.go` change in this PR fixes.

deekay2310:
> @deekay2310 thx for quick review, actually I asked to use your tekon task as reference, when I run this test locally I do need the IBMCLOUD_ACCOUNT.... so not sure if the issue is we are missing it in both Taks (so in secrets format expectation as well)
> 
> Did you test the task?

Ran `mapt ibmcloud ibm-gaudi create` locally with real creds but `IBMCLOUD_ACCOUNT` unset (verified it's not in the env). It got through creating ResourceGroup, VPC, Subnet, SecurityGroup, SSH Key, FloatingIP, and PublicGateway with no issues. Killed it before the instance finished spinning up. The API key alone is enough since it's already scoped to an account.

Grepping the code, `IBMCLOUD_ACCOUNT` is only consumed in the PowerVS paths (piimages.go and power.go), not VPC.

Also the run printed a deprecation warning on the SSH security group rule: property "tcp" is deprecated, use 'protocol', 'code', and 'type' instead, which is exactly what the network.go change in this PR fixes.

adrianriobo:
Yeah, you are right, just tested. I removed the line from doc now it should be good. I will merge once it passes here.

